PR #3, opened 2017-11-01
attempt to build images....

deads2k:
and something went horribly, horribly wrong

I think the original bash for building is somehow broken and it's rippling through everything.  build-images requires rpms and checksums, which require build-cross, which seems to fail.

Really any sane way to build an RPM or image would do for me, but something is going horribly wrong here.

@stevekuznetsov @joelsmith


deads2k:
After speaking with @smarterclayton and @bparees, I can scrap the larger `hack/*` from this repo, do a simple `go build` and create a simple dockerfile from `origin-base` that just copies the binary in for CI testing.

The official cross-build, release, rpm, and rpm-based image need to be solved before shipping, but not before we use this to free up the 3.8 rebase.

bparees:
scrapping hack/* is music to my ears.

deads2k:
closing in favor of https://github.com/openshift/image-registry/pull/4

PR #4, opened 2017-11-01
Simplify building and produce an image

deads2k:
removes hack and replaces it with a stripped down build, makefile, and image creation


images/dockerregistry/Dockerfile.centos7
```diff
@@ -0,0 +1,23 @@
+#
```
bparees: is there a reason we have this file? it's identical to Dockerfile.

bparees:
one question on getting rid of Dockerfile.centos7, otherwise lgtm.

images/dockerregistry/Dockerfile.centos7
deads2k: > is there a reason we have this file? it's identical to Dockerfile.

I have no idea.  I assumed that CI might want it and it didn't cost me anything to update.

Makefile
```diff
@@ -1,209 +1,18 @@
...
-#   test: Run all tests.
-#   run: Run all-in-one server
-#   clean: Clean up.
+all: build
+.PHONY: all
 
-OUT_DIR = _output
-OS_OUTPUT_GOPATH ?= 1
+build:
+	go build -o _output/bin/dockerregistry github.com/openshift/image-registry/cmd/dockerregistry
```
stevekuznetsov: Unclear what  scope this is in but I think we will need:

 - ability to cross-build
 - ability to add `ldflag`s
 - ability to add build tags

images/dockerregistry/Dockerfile.centos7
stevekuznetsov: No need for this for CI

images/dockerregistry/Dockerfile
```diff
@@ -0,0 +1,23 @@
...
+# This is the integrated OpenShift Origin Docker registry. It is configured to
+# publish metadata to OpenShift to provide automatic management of images on push.
+#
+# The standard name for this image is openshift/origin-docker-registry
+#
+FROM openshift/origin-base
+
+ADD bin/dockerregistry /usr/bin/dockerregistry
+
+COPY config.yml ${REGISTRY_CONFIGURATION_PATH}
```
stevekuznetsov: I understand this is pre-existing ... but what sets this var?

Makefile
```diff
@@ -1,209 +1,18 @@
...
-# Example:
-#   make verify
 verify: build
-	# build-tests task has been disabled until we can determine why memory usage is so high
-	{ \
-	hack/verify-gofmt.sh ||r=1;\
-	hack/verify-govet.sh ||r=1;\
-	exit $$r ;\
-	}
+	go test github.com/openshift/image-registry/pkg/...
```
stevekuznetsov: not entirely sure the team working on this wants to give up jUnit output

Makefile
bparees: should not need to cross-build, the registry binary runs in a linux image, it doesn't run on bare metal.

Makefile
bparees: we do not, but i think this is basically a point in time "get something working" and then build back up to full function.

consider it an opportunity to start fresh w/o some of the baggage we've accrued in origin.

images/dockerregistry/Dockerfile.centos7
bparees: so does anyone know why we have it?  (we seem to have it for other images too).

images/dockerregistry/Dockerfile
bparees: yeah i dunno, i don't see anything setting it.  weirdly the value is set a few lines lower in the dockerfile.  I'm not sure how this ever worked?

bparees: (looks like things expect config.yaml to be in / )

bparees: (and it looks like that's where docker dumps the file if the variable is undefined).

so i guess this is "ok" but it's not clear to me if it's intentionally working,  or accidentally.

anyway it's fine to leave it for this PR.

Makefile
stevekuznetsov: @smarterclayton still maintains we need to be able to ship the binary itself

PR #5, opened 2017-11-03
Add integration tests

dmage:


dmage:
@deads2k how can I add new dependencies to the vendor directory using glide? `glide up` and `glide get github.com/RangelReale/osin` change a lot and they break repository.

deads2k:
> @deads2k how can I add new dependencies to the vendor directory using glide? glide up and glide get github.com/RangelReale/osin change a lot and they break repository.

You have a carry: https://github.com/openshift/image-registry/commit/954bdb50033e1697f80db8dc0a5807d8ef2782c8 .  I've been doing

```
glide update --strip-vendor
git checkout vendor/github.com/docker/distribution/vendor
```

until you manage to eliminate your `openshift/vendor`.

deads2k:
```
glide get github.com/RangelReale/osin --strip-vendor
git checkout vendor/github.com/docker/distribution/vendor
```

looks pretty good.

test/integration/dockerregistry_imagelayers_test.go
```diff
@@ -0,0 +1,457 @@
+package integration
```
deads2k: add something to the makefile and preferably to travis to see this work?

dmage: Done.

pkg/cmd/dockerregistry/dockerregistry.go
```diff
@@ -40,21 +40,22 @@ import (
 
 	kubeversion "k8s.io/kubernetes/pkg/version"
 
+	"github.com/openshift/origin/pkg/cmd/server/crypto"
+	"github.com/openshift/origin/pkg/version"
+
+	"github.com/openshift/image-registry/pkg/clientcmd"
```
deads2k: The copy is fine, but I think you should create a package `origin-copy` or some such so that we can keep them all together and see if, when, and how we should be pulling these out into a common library.

bparees:
just to follow up on what @deads2k said, anything we're copying from the origin repo we should nest under pkg/origin-copy so we can keep track of it (and make sure we're periodically syncing it, until we split it out).

dmage:
@bparees @deads2k ok, I'll move it, but it's a modified copy without unused code (to remove dependencies) and with new registry-only features (get the path to kubeconfig from our config file).

deads2k:
> @bparees @deads2k ok, I'll move it, but it's a modified copy without unused code (to remove dependencies) and with new registry-only features (get the path to kubeconfig from our config file).

That approach is fine.  We just need an easy place to find high-value targets across multiple split repos and to assess risk for rebases.  Don't want to lose any.

bparees:
I can confirm i was able to build a docker-registry image using this PR.  I was able to bring up a cluster using the registry image and push to it as well.

deads2k:
Looks like the imports are in a place we can find them now.  I'll leave it with @bparees 

/unassign

miminar:
unit test job seems to be working: `Of 92 tests executed in 15.933s, 92 succeeded, 0 failed, and 0 were skipped. [INFO] hack/test-go.sh exited with code 0 after 00h 01m 42s`
:fireworks:

PR #6, opened 2017-11-07
Incremental RPM build and docker image builds from RPMs

smarterclayton:
```
hack/env make build-images
```

Supports true incremental rebuild on RPM (~10-12s on no-op) so you can have a faster dev loop.

Based on openshift/release#361

@deads2k


smarterclayton:
@bparees your builds, sir.

dmage:
:-1: why do we need this 5k lines of code?

Makefile
```diff
@@ -1,18 +1,124 @@
...
 .PHONY: verify
 
+
+# Verify commit comments.
+#
+# Example:
+#   make verify-commits
+verify-commits:
+	hack/verify-upstream-commits.sh
+.PHONY: verify-commits
```
bparees: why isn't this invoked by the verify target?

Makefile
```diff
@@ -1,18 +1,124 @@
...
+
+# Run unit tests.
+#
+# Args:
+#   WHAT: Directory names to test.  All *_test.go files under these
+#     directories will be run.  If not specified, "everything" will be tested.
+#   TESTS: Same as WHAT.
```
bparees: why do we need both of these?  if we're starting a fresh repo, can we get rid of the some of the crud we've accumulated?

Makefile
```diff
@@ -1,18 +1,124 @@
...
+	rm -rf $(OUT_DIR)
 .PHONY: clean
+
+# Build the cross compiled release binaries
+#
+# Example:
+#   make build-cross
+build-cross:
+	hack/build-cross.sh
+.PHONY: build-cross
```
bparees: why do we need build-cross for the registry?

Makefile
```diff
@@ -1,18 +1,124 @@
...
+#     package, the build will produce executable files under $(OUT_DIR)/local/bin.
+#     If not specified, "everything" will be built.
+#   GOFLAGS: Extra flags to pass to 'go' when building.
+#   TESTFLAGS: Extra flags that should only be passed to hack/test-go.sh
+#
+# Example:
+#   make
+#   make all
+#   make all WHAT=cmd/oc GOFLAGS=-v
+all build:
```
bparees: why doesn't `make all` build rpms and images?  (and run verify for that matter).

having "all" be an alias for "build" seems weird.

hack/OWNERS
```diff
@@ -0,0 +1,2 @@
+reviewers:
```
bparees: dmage legionus miminar bparees

hack/OWNERS
```diff
@@ -0,0 +1,2 @@
...
+approvers:
```
bparees: bparees (and probably @dmage @legionus @miminar )

hack/README.md
```diff
@@ -0,0 +1,34 @@
+# Standard build environment
+
+This is a set of tools for constructing a standard Go build environment for OpenShift.
+
+1. To start, copy the `tools/hack/golang` directory into the root of your repo as `hack`.
+2. Move `package.spec` into your root repository directory as `<REPO_NAME>.spec` (this will be your RPM build)
+3. Move `Makefile` into your root repository directory
```
bparees: why is this not just how the repo comes? i'm confused by this.

hack/README.md
```diff
@@ -0,0 +1,34 @@
...
+4. Update `hack/lib/constants.sh` with the names and variables you care about.
+
+### Run a build 
+
+Creates appropriate binaries in `_output/local/bin/<GOOS>/<GOARCH>`:
+
+```
+$ make
```
bparees: should be make build imho.  "make" (make all) should do everything.

hack/build-base-images.sh
```diff
@@ -0,0 +1,17 @@
+#!/bin/bash
+
+# This script builds the base and release images for use by the release build and image builds.
```
bparees: we're not actually going to do this in this repo are we?  I'd expect base+release images to still come out of origin.

hack/build-cross.sh
```diff
@@ -0,0 +1,95 @@
+#!/bin/bash
```
bparees: don't think we need this.

hack/build-go.sh
```diff
@@ -0,0 +1,21 @@
...
+function cleanup() {
+	return_code=$?
+	os::util::describe_return_code "${return_code}"
+	exit "${return_code}"
+}
+trap "cleanup" EXIT
+
+build_targets=("$@")
+if [[ -z "$@" ]]; then
+	build_targets=("${OS_CROSS_COMPILE_TARGETS[@]}")
```
bparees: what cross compile targets?

hack/build-images.sh
```diff
@@ -0,0 +1,40 @@
+#!/bin/bash
+
+# This script builds all images locally except the base and release images,
+# which are handled by hack/build-base-images.sh.
+
+# NOTE:  you only need to run this script if your code changes are part of
+# any images OpenShift runs internally such as origin-sti-builder, origin-docker-builder,
+# origin-deployer, etc.
```
bparees: comment needs cleanup.

hack/install-tools.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/bash
+STARTTIME=$(date +%s)
+source "$(dirname "${BASH_SOURCE}")/lib/init.sh"
+
+echo "Detected go version: $(go version)"
+
+go get github.com/tools/godep
```
bparees: this repo is using glide, not godep, right?

hack/lib/build/binaries.sh
```diff
@@ -0,0 +1,413 @@
...
+}
+readonly -f os::build::place_bins
+
+# golang 1.5 wants `-X key=val`, but golang 1.4- REQUIRES `-X key val`
+function os::build::ldflag() {
+  local key=${1}
+  local val=${2}
+
+  GO_VERSION=($(go version))
+  if [[ -n $(echo "${GO_VERSION[2]}" | grep -E 'go1.4') ]]; then
```
bparees: we don't care about 1.4 anymore.

bparees:
also doesn't this need an rpm spec file somewhere?

smarterclayton:
This is not cargo culting.  This is the minimum toolchain you'll need to pass the release bar for core code outside of origin.  See the linked PR for an explanation of how this will be kept in sync. 

You have to support multi-arch builds in RPM.  You have to support squashed layered images.  You need to support standard test hooks.  Rolling your own config is not supported.

This is the mechanism by which the people who do all the ugly build work for you can continue to do it in a more sustainable path.  If there are specific parts of the tool chain that you'd like to opt out of, highlight that and we'll either allow it (most people don't need it) or deny it (higher level build needs it, etc).

We are not going to allow/force every team to maintain their own special snowflake Go build tooling.

smarterclayton:
Specfile is in the root.

hack/install-tools.sh
smarterclayton: Yeah, removing, not needed.

hack/lib/build/binaries.sh
smarterclayton: Removing

hack/build-go.sh
smarterclayton: dockerregistry.  Should be testable under any platform, nothing special about it.  IMAGE_COMPILE is for things that only work in images.

hack/build-base-images.sh
smarterclayton: I'm going to remove it for now but we'll probably end up needing a way for this to be reused.  The CI system needs to reuse the latest valid origin-base, even if it's named something else.  Steve and I will have to talk about it.

hack/README.md
smarterclayton: Our convention is make does build.

hack/README.md
smarterclayton: This file is out of date.

hack/OWNERS
smarterclayton: You can do that afterwards

Makefile
smarterclayton: Build cross is what builds everything.  It's the official release binary build.

bparees:
>  If there are specific parts of the tool chain that you'd like to opt out of, highlight that and we'll either allow it (most people don't need it)

cross-platform building.  it's just a waste of time for this repo.

if we want to hack it by hardcoding OS_ONLY_BUILD_PLATFORMS in the appropriate make targets, i can live with that.

Makefile
smarterclayton: No, because other people still need to enforce these conventions in CI tooling.

Makefile
smarterclayton: Because we can only run it when we have a git repo.

Makefile
smarterclayton: Convention today, we can talk about changing it but not here.

smarterclayton:
> cross-platform building. it's just a waste of time for this repo.

No it isn't, because people in this organization develop on other platforms besides Linux.  If you add linux specific code to the registry you break other things.

smarterclayton:
Separate repo doesn't mean "wild west" or "start from scratch".  Conventions are open to change, but slowly over time like all other things.  Teams aren't free to reinvent all wheels here.

hack/build-go.sh
bparees: then dockerregistry should be an IMAGE_COMPILE target imho.  We aren't going to fix it if it's broken on another platform, are we?  nor should pulling in a linux-only godep prevent us from building it.

hack/cherry-pick.sh
```diff
@@ -0,0 +1,88 @@
...
+patch="${patch}/cherry-pick"
+
+if [[ ! -d "${UPSTREAM_REPO_LOCATION}" ]]; then
+  echo "Expected ${UPSTREAM_REPO_LOCATION} to exist" 1>&2
+  exit 1
+fi
+
+lastrev="${NO_REBASE-}"
+if [[ -z "${NO_REBASE-}" ]]; then
+  lastrev="$(go run ${OS_ROOT}/tools/godepversion/godepversion.go ${OS_ROOT}/Godeps/Godeps.json ${repo}/${package})"
```
deads2k: Dependencies on godep should be removed.  I don't see us using long term and we don't have any of the broken out pieces (api, client-go, image-registry) using it.  They're all using glide.

hack/deps
```diff
@@ -0,0 +1,32 @@
...
+set -o nounset
+set -o pipefail
+
+source "$(dirname "${BASH_SOURCE}")/lib/init.sh"
+
+if [[ $# -eq 0 || ! -f "${OS_OUTPUT}/deps" ]]; then
+  echo "Generating dependency graph ..." 1>&2
+  mkdir -p "${OS_OUTPUT}"
+  go list -f '{{.ImportPath}}{{.Imports}}' ./pkg/... ./cmd/... | tr '[]' '  ' | 
+    grep -vE '^github.com/openshift/origin/cmd/(service-catalog|cluster-capacity)' |
```
deads2k: hardcoded `openshift/origin` references (and other commands/repos) need to be removed.

deads2k:
General comments.

As a person using this to stand up a new repo, I'd really like:
 1. a git submodule or some other very structured, easily updateable (to stay current), hard to hack up (enforce via CI) way to get the "standard" library.
 2. an easy, structured, cargo-cultable way to define the points of variability for my build.  It's not just the modification of bash that's hard, it's finding the right bash to modify.  How about a config file that points to various bash files for injecting specific functions I need and a template to fill out.
 3. a script that can check my makefile to see if it CI will accept it.
 4. a way to avoid golang dependencies for tooling. I don't want to have to deconflict vendored dependencies with things like junit.

None of that specifically precludes this, but as an approach to building something I'd be able to consume, those things would really help.

hack/cherry-pick.sh
smarterclayton: I can remove cherry-pick if you don't think you'll need it.

smarterclayton:
As discussed in person:

1. implementation is shared, current approach is a sync managed by someone centrally
2. api is evolving, but agree it would be a small set of overrides / defines per project getting more formal over time

Note that all the golang tooling dependencies require nothing outside the standard lib (or should, which we can enforce with import restrictions).

Re: cargo culting I would expect it to improve over time.

stevekuznetsov:
> a way to avoid golang dependencies for tooling. I don't want to have to deconflict vendored dependencies with things like junit.

This is why my jUnit tool relied ONLY on the stdlib until someone (hmm!!!) added more deps. The tools we write should be tiny and self-contained.

stevekuznetsov:
@deads2k i would 100% agree with your points except for the go tooling point, addressed in my previous comment.

bparees:
constants.sh and rpm spec look fine.

hack/cherry-pick.sh
```diff
@@ -0,0 +1,88 @@
+#!/bin/bash
+
+# See HACKING.md for usage
```
dmage: Which HACKING.md?

hack/cherry-pick.sh
```diff
@@ -0,0 +1,88 @@
...
+source "$(dirname "${BASH_SOURCE}")/lib/init.sh"
+
+repo="${UPSTREAM_REPO:-k8s.io/kubernetes}"
```
dmage: It's not our upstream.

hack/lib/cleanup.sh
```diff
@@ -0,0 +1,191 @@
...
+#  None
+# Returns:
+#  None
+function os::cleanup::truncate_large_logs() {
+	local max_file_size="200M"
+	os::log::info "[CLEANUP] Truncating log files over ${max_file_size}"
+	for file in $(find "${ARTIFACT_DIR}" "${LOG_DIR}" -type f -name '*.log' \( -size +${max_file_size} \)); do
+		mv "${file}" "${file}.tmp"
+		echo "LOGFILE TOO LONG ($(du -h "${file}.tmp")), PREVIOUS BYTES TRUNCATED. LAST ${max_file_size} OF LOGFILE:" > "${file}"
+		tail -c ${max_file_size} "${file}.tmp" >> "${file}"
```
dmage: Does it supposed to work on macOS? `-c 200M` is a GNU extension.

hack/lib/cmd.sh
```diff
@@ -0,0 +1,619 @@
...
+# os::cmd::internal::run_collecting_output runs the command given, piping stdout and stderr into
+# the given files, and returning the exit code of the command
+function os::cmd::internal::run_collecting_output() {
+	local cmd=$1
+
+	local result=
+	$( eval "${cmd}" 1>>"${os_cmd_internal_tmpout}" 2>>"${os_cmd_internal_tmperr}" ) || result=$?
+	local result=${result:-0} # if we haven't set result yet, the command succeeded
+
+	return "${result}"
```
dmage:
```
function os::cmd::internal::run_collecting_output() {
    local cmd=$1
    eval "${cmd}" 1>>"${os_cmd_internal_tmpout}" 2>>"${os_cmd_internal_tmperr}"
}
```

hack/lib/cmd.sh
```diff
@@ -0,0 +1,619 @@
...
+# os::cmd::internal::seconds_since_epoch returns the number of seconds elapsed since the epoch
+# with milli-second precision
+function os::cmd::internal::seconds_since_epoch() {
+	local ns=$(date +%s%N)
+	# if `date` doesn't support nanoseconds, return second precision
+	if [[ "$ns" == *N ]]; then
+		date "+%s.000"
+		return
+	fi
+	echo $(bc <<< "scale=3; ${ns}/1000000000")
```
dmage: Why do you need echo?

hack/lib/constants.sh
```diff
@@ -0,0 +1,139 @@
...
+readonly OS_OUTPUT_SCRIPTPATH="${OS_OUTPUT_SCRIPTPATH:-"${OS_BASE_OUTPUT}/scripts"}"
+
+readonly OS_OUTPUT_SUBPATH="${OS_OUTPUT_SUBPATH:-${OS_OUTPUT_BASEPATH}/local}"
+readonly OS_OUTPUT="${OS_ROOT}/${OS_OUTPUT_SUBPATH}"
+readonly OS_OUTPUT_RELEASEPATH="${OS_OUTPUT}/releases"
+readonly OS_OUTPUT_RPMPATH="${OS_OUTPUT_RELEASEPATH}/rpms"
+readonly OS_OUTPUT_BINPATH="${OS_OUTPUT}/bin"
+readonly OS_OUTPUT_PKGDIR="${OS_OUTPUT}/pkgdir"
+
+readonly OS_GOFLAGS_TAGS="include_gcs include_oss containers_image_openpgp"
```
dmage: containers_image_openpgp is not present in our repo

hack/lib/test/junit.sh
```diff
@@ -0,0 +1,204 @@
...
+#  - NUM_OS_JUNIT_SUITES_IN_FLIGHT
+# Arguments:
+#  - 1: the suite name that is starting
+# Returns:
+#  - increment NUM_OS_JUNIT_SUITES_IN_FLIGHT
+function os::test::junit::declare_suite_start() {
+    local suite_name=$1
+    local num_suites=${NUM_OS_JUNIT_SUITES_IN_FLIGHT:-0}
+
+    echo "=== BEGIN TEST SUITE github.com/openshift/origin/test/${suite_name} ===" >> "${JUNIT_REPORT_OUTPUT:-/dev/null}"
```
dmage: openshift/origin

hack/lib/test/junit.sh
```diff
@@ -0,0 +1,204 @@
...
+function os::test::junit::internal::generate_report() {
+    local report_type="$1"
+    os::util::ensure::built_binary_exists 'junitreport'
+
+    local report_file
+    report_file="$( mktemp "${ARTIFACT_DIR}/${report_type}_report_XXXXX" ).xml"
+    os::log::info "jUnit XML report placed at $( os::util::repository_relative_path ${report_file} )"
+    junitreport --type "${report_type}"             \
+                --suites nested                     \
+                --roots github.com/openshift/origin \
```
dmage: openshift/origin

hack/lib/util/find.sh
```diff
@@ -0,0 +1,73 @@
...
+#  - GOPATH
+# Arguments:
+#  - 1: binary name
+# Returns:
+#  - location of the binary
+function os::util::find::gopath_binary() {
+	local binary_name="$1"
+
+	local old_ifs="${IFS}"
+	IFS=":"
```
dmage: Just `local IFS=:`?

hack/move-upstream.sh
```diff
@@ -0,0 +1,62 @@
+#!/bin/bash
+
+# See HACKING.md for usage
+# To apply all the kube UPSTREAM patches to a kubernetes git directory, you can
```
dmage: I guess we don't need this.

hack/test-go.sh
```diff
@@ -0,0 +1,192 @@
...
+    go test -i ${gotest_flags} ${test_packages}
+    go test ${gotest_flags} ${test_packages} 2>"${test_error_file}" | tee "${JUNIT_REPORT_OUTPUT}"
+
+    test_return_code="${PIPESTATUS[0]}"
+
+    set -o pipefail
+
+    if [[ -s "${test_error_file}" ]]; then
+        os::log::warning "\`go test\` had the following output to stderr:
+$( cat "${test_error_file}") "
```
dmage: When is this file deleted?

image-registry.spec
```diff
@@ -0,0 +1,130 @@
...
+
+%files
+%doc README.md
+%license LICENSE
+%{_bindir}/dockerregistry
+
+%pre
+
+%changelog
+* Mon Nov 06 2017 Anonymous <anon@nowhere.com> 0.0.1
```
dmage: Anon?

deads2k:
> @deads2k i would 100% agree with your points except for the go tooling point, addressed in my previous comment.

Can you think of any reason a user of the tools (like image-registry) would need to customize it?  The tool is useful and self-contained.  Regardless of its dependency tree, it seems like we'd want to depend on it as a binary dependency (like godep or glide), but not as a source dependency.

tools/import-verifier/import-verifier.go
```diff
@@ -0,0 +1,359 @@
...
+	"io"
+	"io/ioutil"
+	"log"
+	"os"
+	"os/exec"
+	"strings"
+)
+
+const (
+	rootPackage = "github.com/openshift/origin"
```
dmage: openshift/origin

stevekuznetsov:
@deads2k how would we publish and deliver the tool as a binary?

deads2k:
> @deads2k how would we publish and deliver the tool as a binary?

It works for things like `glide`.  I think you just have to publish a release of the tool on github.

stevekuznetsov:
Does it also work for `dep`? I don't necessarily want us to be locked into a specific dependency manger

deads2k:
> Does it also work for dep? I don't necessarily want us to be locked into a specific dependency manager

Yes.  Most tools like that are go gettable in a way that leaves you with a binary in GOPATH/bin

stevekuznetsov:
Depending on binaries then SGTM

stevekuznetsov:
We could break those tools out into separate repos I guess so we can release them independently (or can we do it from `openshift/origin`?)

deads2k:
> We could break those tools out into separate repos I guess so we can release them independently (or can we do it from openshift/origin?)

I don't think keeping them in openshift/origin is appropriate.  I would just create a separate repo for them.

smarterclayton:
The tools are now in openshift/release

stevekuznetsov:
@smarterclayton is this ready? can I build the image without this PR for the CI for this repo?

stevekuznetsov:
To reiterate what I told David on the failure -- we are expecting it to fail. We need to update the base AMI to bake in the location for this repo. Base AMI is currently blocked on some e2es -- I have posted about that for a while in aos-devel. If it doesn't look like we can resolve the e2es issue in the AMI build I will investigate some other hack to get this unblocked.

stevekuznetsov:
Tests should be unblocked now

/test e2e

stevekuznetsov:
Bad `make` target in the job, fixed in https://github.com/openshift/aos-cd-jobs/pull/843

/test e2e

stevekuznetsov:
@bparees @smarterclayton @deads2k the PR test ran -- built the image from this repo, tagged it into the Origin SHA, used it in an Ansible install and the e2es all passed. Is this PR ready to merge? After that happens, I can update the Origin e2es to compile registry from this repo

bparees:
it's ok with me.

deads2k:
> @bparees @smarterclayton @deads2k the PR test ran -- built the image from this repo, tagged it into the Origin SHA, used it in an Ansible install and the e2es all passed. Is this PR ready to merge? After that happens, I can update the Origin e2es to compile registry from this repo

I don't know what it does, but we need to unblock the rebase and clayton says its ok.  No one else is planning to step up to do something cleaner.  Merging this to unstick us.

PR #7, opened 2017-11-09
yaml-enable the registry rate limiting config

bparees:


bparees:
@smarterclayton I assume this won't mess up what you're doing if we merge it.  I think this is the only delta from origin that didn't make the copy into image-registry.

PR #19, opened 2017-11-28
Use docker client in the integration tests

dmage:
/assign @legionus


hack/lib/constants.sh
```diff
@@ -7,7 +7,7 @@ readonly OS_GO_PACKAGE=github.com/openshift/image-registry
 readonly OS_BUILD_ENV_GOLANG="${OS_BUILD_ENV_GOLANG:-1.8}"
 readonly OS_BUILD_ENV_IMAGE="${OS_BUILD_ENV_IMAGE:-openshift/origin-release:golang-${OS_BUILD_ENV_GOLANG}}"
 readonly OS_REQUIRED_GO_VERSION="go1.8"
-readonly OS_BUILD_ENV_WORKINGDIR="/go/${OS_GO_PACKAGE}"
+readonly OS_BUILD_ENV_WORKINGDIR="/go/src/${OS_GO_PACKAGE}"
```
dmage: @stevekuznetsov @kargakis ptal

stevekuznetsov: Why did we need this? /cc @smarterclayton

stevekuznetsov: IF we all need this, it should get percolated up

dmage: We need this because we want our package in GOPATH.

stevekuznetsov: Can you please push this to the `openshift/release` example? That is authoritative.

PR #22, opened 2017-11-29
bump(*)

bparees:
the point of this PR is to update glide.yaml to use the new image-registry-release-2.6.2 branch in openshift/docker-distribution.

everything else is fall out of running glide update after making that change....  so it needs validation.  In particular we should not have seen any changes to the docker-distribution vendor dir itself, and yet there is one, so we need to understand where that came from.

update: that has been resolved by merging the carry into the fork.

### DIFF
--- a/glide.lock
+++ b/glide.lock
@@ -1,5 +1,5 @@
-hash: b7794dd8cdab3499c9dc242dfaf162b44ac551d7f37c0c07b8698d66c552d649
-updated: 2017-11-28T16:05:27.008433494+01:00
+hash: 68e66e396fe1d148b422e5175c936b9e0201eb4693ee5f841a484c957c1bea1e
+updated: 2017-11-30T13:44:53.963922749-05:00
 imports:
 - name: bitbucket.org/ww/goautoneg
   version: 75cd24fc2f2c2a2088577d12123ddee5f54e0675
@@ -83,7 +83,7 @@ imports:
   - oss
   - util
 - name: github.com/docker/distribution
-  version: 1e2bbed6e09c6c8047f52af965a6e301f346d04e
+  version: 588ea2c42a2ebb5c06cd01dbeabfe60e4a7075df
   repo: git@github.com:openshift/docker-distribution
   subpackages:
   - configuration
@@ -389,6 +389,9 @@ imports:
   - pkg/image/generated/clientset/scheme
   - pkg/image/generated/clientset/typed/image/v1
   - pkg/image/generated/clientset/typed/image/v1/fake
+  - pkg/image/generated/internalclientset
+  - pkg/image/generated/internalclientset/scheme
+  - pkg/image/generated/internalclientset/typed/image/internalversion
   - pkg/image/importer
   - pkg/image/importer/dockerv1client
   - pkg/image/util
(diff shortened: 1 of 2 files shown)


vendor/github.com/docker/distribution/configuration/parser.go
```diff
@@ -220,7 +220,7 @@ func (p *Parser) overwriteStruct(v reflect.Value, fullpath string, path []string
 		}
 	case reflect.Ptr:
 		if field.IsNil() {
-			field.Set(reflect.New(field.Type().Elem()))
+			field.Set(reflect.New(sf.Type))
```
bparees: i'm not sure why this change is being picked up unless we were carrying a patch in our vendor dir that is not in openshift/docker-distribution ?

deads2k:
@legionus @dmage @miminar did you guys find a way to eliminate `UPSTREAM: docker/distribution: <drop>: pull vendor deps` now that you're no longer in a repo trying to build `kubelet`?  That would be great, but I didn't notice it go by so I may have given @bparees out of date advice.

vendor/github.com/docker/distribution/configuration/parser.go
bparees: it looks like the change came from 7746110b40a1689628d79c8ba35822746ffd3160 

so we need to get 7746110b40a1689628d79c8ba35822746ffd3160 into openshift/docker-distribution#image-registry-release-2.6.2

bparees: pr here: https://github.com/openshift/docker-distribution/pull/2

dmage:
@deads2k yes, `glide up -v` and one patch for distribution should be enough.

bparees:
thanks @dmage, i dropped the carry.  after the change merges into docker-distribution i'll rerun glide up on this PR.

vendor/github.com/docker/distribution/configuration/parser.go
legionus: > i'm not sure why this change is being picked up unless we were carrying a patch in our vendor dir that is not in openshift/docker-distribution ?

@mfojtik told me that changes in the `vendor` directory should be made in our repository and then transferred to the `openshift/docker-distribution` repository by some script.

Before that, I was opening PR: https://github.com/openshift/docker-distribution/pull/1

bparees: well barring more details on who's implementing that script, i'm going to proceed with the manual approach.

bparees:
ok i think this is finally doing what i'd expect.  Our docker-distribution vendor dep now comes out of the openshift/docker-distribution:image-registry-release-2.6.2 fork+branch.

Going forward, we should submit carries to that fork/branch and then vendor back to here.

@dmage @legionus @deads2k ptal

legionus:
>  Our docker-distribution vendor dep now comes out of the openshift/docker-distribution:image-registry-release-2.6.2 fork+branch.

@bparees Only you will have permissions to merge there ?

bparees:
> @bparees Only you will have permissions to merge there ?

any openshift project committer (there are about 22 last i knew).  Certainly we can consider expanding that as well, or perhaps we'll get the tooling that @mfojtik alluded to.

PR #23, opened 2017-12-04
Use a generic wrapper for audit and metrics

dmage:


pkg/dockerregistry/server/wrapped/zz_funcname_test.go
```diff
@@ -0,0 +1,68 @@
...
+import (
+	"fmt"
+	"reflect"
+	"strings"
+	"testing"
+
+	"github.com/docker/distribution/context"
+)
+
+func TestFuncName(t *testing.T) {
```
bparees: would it make more sense to call this a wrapper test?

pkg/dockerregistry/server/wrapped/zz_wrapper.go
```diff
@@ -0,0 +1,7 @@
+package wrapped
```
bparees: why "zz_wrapper.go" and not just "wrapper.go"?

miminar:
LGTM as long as Ben's questions are resolved.

pkg/dockerregistry/server/wrapped/zz_funcname_test.go
dmage: renamed

pkg/dockerregistry/server/wrapped/zz_wrapper.go
dmage: renamed

dmage:
@legionus done, the error middleware is splitted in two pieces.

PR #28, opened 2017-12-13
Supermiddleware

dmage:
The intention of the changes is to move all magic for middlewares into a single package.


dmage:
flake openshift/origin#17786
/retest

pkg/dockerregistry/server/blobdescriptorservice.go
```diff
@@ -34,61 +30,50 @@ func (b ByGeneration) Less(i, j int) bool { return b[i].Generation > b[j].Genera
...
+func (f blobDescriptorServiceFactoryFunc) BlobAccessController(svc distribution.BlobDescriptorService) distribution.BlobDescriptorService {
+	return f(svc)
 }
 
 type blobDescriptorService struct {
 	distribution.BlobDescriptorService
+	repo *repository
+}
+
+func (r *repository) BlobDescriptorService(svc distribution.BlobDescriptorService) distribution.BlobDescriptorService {
```
legionus: Please move this function to `repository.go`

pkg/dockerregistry/server/repository.go
```diff
@@ -133,18 +132,16 @@ func (r *repository) Manifests(ctx context.Context, options ...distribution.Mani
 	ms = &manifestService{
 		repo:          r,
 		manifests:     ms,
-		acceptschema2: r.config.acceptSchema2,
+		acceptschema2: r.app.extraConfig.Compatibility.AcceptSchema2,
 	}
 
-	if r.config.pullthrough {
+	if r.app.extraConfig.Pullthrough.Enabled {
```
legionus: Please use `r.app.extraConfig` instead of `r.enabledMetrics` as well.

pkg/dockerregistry/server/auth.go
```diff
@@ -89,7 +89,8 @@ var (
 	ErrUnsupportedResource = errors.New("unsupported resource")
 )
 
-func (app *App) newAccessController(authcfg *configuration.Auth) (registryauth.AccessController, error) {
+func (app *App) Auth(options map[string]interface{}) (registryauth.AccessController, error) {
+	authcfg := app.extraConfig.Auth
 	tokenRealm, err := configuration.TokenRealm(authcfg.TokenRealm)
```
legionus:
```diff
- authcfg := app.extraConfig.Auth
- tokenRealm, err := configuration.TokenRealm(authcfg.TokenRealm)
+ tokenRealm, err := configuration.TokenRealm(app.extraConfig.Auth.TokenRealm)
```
?

pkg/dockerregistry/server/manifestservice.go
```diff
@@ -68,7 +68,7 @@ func (m *manifestService) Get(ctx context.Context, dgst digest.Digest, options .
 	ref := imageapi.DockerImageReference{
 		Namespace: m.repo.namespace,
 		Name:      m.repo.name,
-		Registry:  m.repo.config.registryAddr,
+		Registry:  m.repo.app.extraConfig.Server.Addr,
```
legionus: Can we rename `extraConfig` to `config` ?

pkg/dockerregistry/server/blobdescriptorservice.go
dmage: I'd like to move as few functions as possible in this PR.

pkg/dockerregistry/server/manifestservice.go
dmage: It's a good idea for follow ups.

legionus:
@dmage can you add `doc.go` for supermiddleware ?

pkg/dockerregistry/server/auth.go
dmage: done

pkg/dockerregistry/server/repository.go
dmage: done

dmage:
@legionus doc.go - done

pkg/dockerregistry/server/supermiddleware/registry.go
```diff
@@ -0,0 +1,78 @@
...
+
+func (reg *registry) Repository(ctx context.Context, named reference.Named) (distribution.Repository, error) {
+	repo, err := reg.Namespace.Repository(ctx, named)
+	if err != nil {
+		return repo, err
+	}
+	return reg.inst.Repository(ctx, repo, false)
+}
+
+// HackRegistry must not be used.
```
bparees: so why is it here? can it be moved into a _test.go file?

dmage:
@bparees the main problem is BlobDescriptorService (it determines whether the blob exists in the repository). We can register only one implementation, which doesn't get any parameters.

https://github.com/openshift/image-registry/blob/c540fa0224d2028b9633a147c283fbf60d371ba9/pkg/dockerregistry/server/blobdescriptorservice.go#L37-L42

However we need to access the image stream from this service. And we can get it from the context.

https://github.com/openshift/image-registry/blob/c540fa0224d2028b9633a147c283fbf60d371ba9/pkg/dockerregistry/server/blobdescriptorservice.go#L62-L67
https://github.com/openshift/image-registry/blob/c540fa0224d2028b9633a147c283fbf60d371ba9/pkg/dockerregistry/server/blobdescriptorservice.go#L90

It's pretty ugly, but I don't see any better options. But how do we put the repository into the context? Basically we have `withRepository(ctx, r.repo)` all around the code. And recently I replaced all such calls with my generic wrapper.

https://github.com/openshift/image-registry/blob/c540fa0224d2028b9633a147c283fbf60d371ba9/pkg/dockerregistry/server/withrepository.go

So, the problem is solved, right? No, it isn't. :(

If BlobStore.Create(ctx, ...) is called with the WithMountFrom(src) option, we should be very careful with contexts, the upstream code may call Stat on the source repository with the context from the destination repository. We have a workaround for this

https://github.com/openshift/image-registry/blob/c540fa0224d2028b9633a147c283fbf60d371ba9/pkg/dockerregistry/server/pendingerrors.go#L56-L59
https://github.com/openshift/image-registry/blob/c540fa0224d2028b9633a147c283fbf60d371ba9/pkg/dockerregistry/server/pendingerrors.go#L130-L151

But, unfortunately, it has a bug. The context isn't modified before calling Stat.

Back to the PR: writing middlewares for the blob descriptor service should be less intriguing and detective, so now we have the entire implementation of the blob descriptor service middleware in the context. And that allowed me to cover it with tests. Though, this PR is just refactoring, bug fixes will be made as follow ups.

pkg/dockerregistry/server/supermiddleware/registry.go
dmage: It's a huge hack and should be removed, it's here to make a private variable accessible from another module. We gave too much tasks to the type `*repository`. It implements distribution.Repository and it's responsible for getting data from the image stream. But, for example, pullthroughManifestService wants `*repository` and uses

  * m.repo.namespace
  * m.repo.name
  * m.repo.getImageOfImageStream
  * m.repo.rememberLayersOfManifest
  * m.repo.registryOSClient
  * m.repo.imageStreamGetter

In other words, it wants to fill cache and to get ImageStreamImage. But right now we can't get imageStreamGetter and cache without constructing entire *repository.

Splitting will be almost mechanical changes, but it will be pretty big and I don't want to mix it with this PR.

legionus:
> What problem is solved by moving them into a single package?

@bparees This is a very important change that corrects the current architecture. Our code will become more maintainable. Also it will allow us to change the caching.

bparees:
@dmage thanks for the summary, that was very helpful.

PR #36, opened 2017-12-21
Refactor crossmount test

dmage:


bparees:
aren't there cases where cross-mounting between projects is valid (when you have read permission to the other project)?  do we have tests for those?

dmage:
Yes, there are. I need to amend the PR.

miminar:
So far it looks good to me. Good job.

PR #37, opened 2018-01-02
Extract imageStream object

dmage:


pkg/dockerregistry/server/imagestream.go
```diff
@@ -0,0 +1,27 @@
...
+
+type imageStream struct {
+	namespace string
+	name      string
+
+	registryOSClient client.Interface
+
+	// cachedImages contains images cached for the lifetime of the request being handled.
+	cachedImages map[digest.Digest]*imageapiv1.Image
+	// cachedImageStream stays cached for the entire time of handling signle repository-scoped request.
```
bparees: s/signle/single/

pkg/dockerregistry/server/repository.go
```diff
@@ -216,9 +213,9 @@ func (r *repository) BlobDescriptorService(svc distribution.BlobDescriptorServic
 }
 
 // createImageStream creates a new image stream corresponding to r and caches it.
```
bparees: godoc needs updating

pkg/dockerregistry/server/blobdescriptorservice.go
```diff
@@ -127,51 +127,51 @@ func imageStreamHasBlob(r *repository, dgst digest.Digest) bool {
...
 
 // imageHasBlob returns true if the image identified by imageName refers to the given blob. The image is
 // fetched. If requireManaged is true and the image is not managed (it refers to remote registry), the image
-// will not be processed. Fetched image will update local cache of blobs -> repositories with (blobDigest,
-// cacheName) pairs.
+// will not be processed.
 func imageHasBlob(
-	r *repository,
+	ctx context.Context,
+	imageStream *imageStream,
```
legionus: Can we do something like this ?

```go
image, err := imageStream.getImage(r.ctx, digest.Digest(tagEvent.Image))
if err != nil {
	...
	return logFound(false)
}
if !r.app.config.Pullthrough.Enabled && !isImageManaged(image) {
	...
	return logFound(false)
￼}
if imageHasBlob(r.ctx, image, dgst.String()) {
tagName := event2Name[tagEvent]
	context.GetLogger(r.ctx).Debugf("blob found under istag %s:%s in image %s", r.imageStream.Reference(), tagName, tagEvent.Image)
	imageStream.rememberLayersOfImage(r.ctx, image, repoCacheName)
￼  	return logFound(true)
}
```

pkg/dockerregistry/server/repository.go
```diff
@@ -323,76 +319,56 @@ func (r *repository) getImageOfImageStream(dgst digest.Digest) (*imageapiv1.Imag
...
+	} else {
+		manifest, err := NewManifestFromImage(image)
+		if err != nil {
+			context.GetLogger(ctx).Errorf("cannot remember layers of image %s: %v", image.Name, err)
+			return
+		}
+		for _, ref := range manifest.References() {
+			var desc *distribution.Descriptor
+			if isImageManaged(image) {
+				desc = &ref
```
legionus: `desc = &refs[i]` ?

pkg/dockerregistry/server/manifesthandler.go
```diff
@@ -36,36 +36,25 @@ type ManifestHandler interface {
...
 
-// NewManifestHandlerFromImage creates a new manifest handler for a manifest stored in the given image.
-func NewManifestHandlerFromImage(repo *repository, image *imageapiv1.Image) (ManifestHandler, error) {
-	var (
-		manifest distribution.Manifest
-		err      error
-	)
-
+// NewManifestFromImage creates a manifest for a manifest stored in the given image.
+func NewManifestFromImage(image *imageapiv1.Image) (distribution.Manifest, error) {
```
legionus:
```go
if len(image.DockerImageManifest) == 0 {
	return nil, fmt.Errorf("unable to create manifest for %s image: empty field", image.Name)
}
```
maybe better to add check to avoid unmarshal errors ?

dmage: It's never called with a such precondition.

pkg/dockerregistry/server/repository.go
dmage: fixed

pkg/dockerregistry/server/manifesthandler.go
dmage: Though, fixed.

pkg/dockerregistry/server/blobdescriptorservice.go
dmage: done

PR #44, opened 2018-01-10
switch junit tools dependency from origin

bparees:


bparees:
@dmage @legionus ptal.

PR #49, opened 2018-01-17
Fix performance regression

legionus:
The docker/distribution has its own blobstater. All our middlewaries are ignored.
So when ServeBlob called, some of the storage requests will not go to our cache.

### DIFF
--- a/pkg/dockerregistry/server/configuration/configuration_test.go
+++ b/pkg/dockerregistry/server/configuration/configuration_test.go
@@ -682,7 +682,8 @@ openshift:
 	}
 }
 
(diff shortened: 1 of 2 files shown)


pkg/dockerregistry/server/configuration/configuration_test.go
```diff
@@ -682,7 +682,7 @@ openshift:
 	}
 }
 
-func TestDisableInmemoryCache(t *testing.T) {
+func testDisableInmemoryCache(t *testing.T) {
```
miminar: It should probably have the same TODO comment.

PR #50, opened 2018-01-17
Stick to specific version of OpenShift Origin container

dmage:


dmage:
https://openshift-gce-devel.appspot.com/build/origin-ci-test/pr-logs/pull/openshift_image-registry/49/test_pull_request_image_registry_integration/130/?log#log

pkg/testframework/versions.go
```diff
@@ -0,0 +1,4 @@
+package testframework
+
+// originImageRef is the Docker image name of the OpenShift Origin container.
+const originImageRef = "docker.io/openshift/origin:v3.9.0-alpha.2"
```
bparees: we're going to have to keep updating this w/ every release?  doesn't seem like the right solution.

dmage: We definitely don't want to use origin's unstable channel.

bparees: no?  if someone is putting code into origin master that breaks the registry, we probably do want to know about it.  (an if done properly, origin's CI merge testing should be setup to disallow such merges)

legionus: @bparees Ok. Right now we faced that the code that worked in the origin before stopped working. What should we do next?

bparees: fix our code or fix the origin code, depending on which one is wrong...   if you're saying the current image-registry:latest image can't work w/ the current origin:latest, that should be considered a p0.

(and again, moving forward we need to get a CI process in place that doesn't allow merges that create that situation.  Merges to origin need to be tested w/ latest image-registry and vice versa)

bparees: now, when we start cutting branches of this repo (3.9, 3.10, etc) there's going to be an additional challenge.  image-registry/v3.9 should be tested w/ the latest origin:v3.9 image, and so on.  Again this is a challenge we need to solve as an organization for a number of repositories.

bparees: @legionus also can you elaborate on this:

> Right now we faced that the code that worked with the origin before stopped working.

what broke and how/why did it break?

dmage: @bparees it's hard to tell what is wrong.

https://github.com/openshift/image-registry/blob/4e128e42af9068986de54f1ac152ef250e0c0e6f/pkg/testframework/project.go#L27-L39

This doesn't give `adminUser` permissions to pull images. Or, to be more precise, now we get:

> User \"admin\" cannot get imagestreams/layers.image.openshift.io in project \"namespace\"

Why should such changes in origin block our merge queue?

bparees: > Why should such changes in origin block our merge queue?

because we have to deliver image-registry code that works w/ the latest origin apis.  If something changed/broke about those apis, we can't just ignore it by going to an older version of origin.  It's not like we can stay on the origin 3.9-alpha.2 image forever in our repo just to keep our tests working.  We need to fix our tests (or if origin actually broke something, get the appropriate team involved).  I would start by enlisting @mfojtik and the master team on this.

dmage: @deads2k openshift/origin#17976 - does it change policies for admin?

bparees:
@dmage please open an issue to track reverting this when the origin change is sorted out.

PR #57, opened 2018-01-26
Check if mirrored image is available in offline

dmage:


dmage:
PROVISION CLOUD RESOURCES [00h 10m 20s] failed
/retest

Issue #58, opened 2018-01-28
REGISTRY_OPENSHIFT_SERVER_ADDR env var does not appear to work


bparees:
the registry does not appear to actually respect this value properly:

```
$ oc logs docker-registry-1-knwnf -n default | grep URL

time="2018-01-28T03:00:12.310853118Z" level=info msg="Using \"172.30.183.222:5000\" as Docker Registry URL" go.version=go1.9.2 instance.id=a82b7ca8-0b55-4567-92ef-2c1744e6074a 

$ oc get pod docker-registry-1-knwnf -o yaml | grep -A 2 ADDR 
--
    - name: REGISTRY_OPENSHIFT_SERVER_ADDR
      value: docker-registry.default.svc:5000
```

bparees:
When this is fixed, we should un-revert https://github.com/openshift/openshift-ansible/pull/6913

legionus:
```
$ oc cluster up --version='latest'
$ oc version
oc v3.9.0-alpha.3+9ab7a71-232-dirty
kubernetes v1.9.1+a0ce1bc657
features: Basic-Auth GSSAPI Kerberos SPNEGO

Server https://127.0.0.1:8443
openshift v3.9.0-alpha.3+9ab7a71-232
kubernetes v1.9.1+a0ce1bc657

$ oc login -u system:admin

$ oc rsh dc/docker-registry cat /proc/1/environ |tr '\0' '\n' |egrep '(DOCKER_REGISTRY_SERVICE_HOST|DOCKER_REGISTRY_SERVICE_POST|DOCKER_REGISTRY_URL|REGISTRY_MIDDLEWARE_REPOSITORY_OPENSHIFT_DOCKERREGISTRYURL|OPENSHIFT_DEFAULT_REGISTRY|REGISTRY_MIDDLEWARE_REPOSITORY_OPENSHIFT_DOCKERREGISTRYURL)'
DOCKER_REGISTRY_SERVICE_HOST=172.30.1.1
```

I do not know how you created the cluster, but in the cluster that is created by `oc cluster up`, a variable `DOCKER_REGISTRY_SERVICE_HOST` is set.

```
$ oc env -n default dc/docker-registry REGISTRY_OPENSHIFT_SERVER_ADDR=foobar

[...]

PR #59, opened 2018-01-29
kube env variables has lower priority than configuration

legionus:
Fixes https://github.com/openshift/image-registry/issues/58

### DIFF
--- a/pkg/dockerregistry/server/configuration/configuration_test.go
+++ b/pkg/dockerregistry/server/configuration/configuration_test.go
@@ -572,6 +572,64 @@ storage:
     rootdirectory: /registry
 openshift:
   version: 1.0
+`
+	type env struct {
+		name, value string
+	}
+	testCases := []struct {
+		setenv   []env
+		expected string
+	}{
+		{
+			setenv: []env{
+				{name: "DOCKER_REGISTRY_SERVICE_HOST", value: "DOCKER_REGISTRY_SERVICE_HOST"},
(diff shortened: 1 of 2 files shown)


pkg/dockerregistry/server/configuration/configuration_test.go
```diff
@@ -583,19 +581,79 @@ openshift:
...
 		{
 			setenv: []env{
 				{name: "DOCKER_REGISTRY_SERVICE_HOST", value: "DOCKER_REGISTRY_SERVICE_HOST"},
 				{name: "DOCKER_REGISTRY_SERVICE_PORT", value: "DOCKER_REGISTRY_SERVICE_PORT"},
 			},
-			expected: "DOCKER_REGISTRY_SERVICE_HOST:DOCKER_REGISTRY_SERVICE_PORT",
+			expected: "from-config",
+		},
+		{
+			setenv:   []env{{name: "REGISTRY_OPENSHIFT_SERVER_ADDR", value: "REGISTRY_OPENSHIFT_SERVER_ADDR"}},
```
bparees: this test would be more thorough if it also set the DOCKER_REGISTRY_SERVICE_XXX variables (to ensure they are ignored).

legionus: @bparees I'm not following. These variables will be set on the previous iteration. Each case in this test adds variables, but does not unset them.

What do you mean ?

bparees: ah, I did not realize that, I assumed each TC was independent.

PR #60, opened 2018-01-29
switch back to using latest origin version

bparees:


pkg/testframework/project.go
```diff
@@ -25,13 +25,15 @@ func CreateProject(t *testing.T, clientConfig *rest.Config, namespace string, ad
...
 		ObjectMeta: metav1.ObjectMeta{
-			Name: "admin",
+			Name: "admin2",
 		},
 		UserNames: []string{adminUser},
+		// this roleref is needed since https://github.com/openshift/origin/pull/17976
+		// seems like maybe the aggregator isn't running for us.
 		RoleRef: corev1.ObjectReference{
-			Name: "admin",
+			Name: "system:openshift:aggregate-to-admin",
```
bparees: @deads2k seems like the role aggregration isn't happening when we run a master w/ basic defaults.  Or maybe it takes longer than we're giving it to happen?  Care to speculate on either possibility?

this is our startup logic, where imageref is the origin image:
```
	resp, err := cli.ContainerCreate(ctx, &container.Config{
		Image:      originImageRef,
		Entrypoint: []string{"/bin/sh"},
		Cmd: []string{"-ec", `
			openshift start master --write-config=/var/lib/origin/openshift.local.config/master
			sed -i'' -e '/- domainName:/d' /var/lib/origin/openshift.local.config/master/master-config.yaml
			exec openshift start master --config=/var/lib/origin/openshift.local.config/master/master-config.yaml
		`},
	}, nil, nil, "")
```

bparees:
@dmage ptal

pkg/testframework/project.go
dmage: The Origin integration tests wait for role aggregation by their own: https://github.com/openshift/origin/blob/9ab7a71de8bde9d863990dab10a758a0cd90e811/test/util/server/server.go#L471-L510
Don't we want to use the same approach?

bparees: sure, that is better, wasn't aware of it.  will update.

bparees:
@dmage updated w/ your suggestion, thanks!

deads2k:
The roles are aggregated via controllers. That doesn't happen until your controllers are started.

bparees:
> The roles are aggregated via controllers. That doesn't happen until your controllers are started.

The controllers are all starting.  The issue was simply that we had to wait for the aggregation to occur before starting our tests (as noted by @dmage, that's what other tests are doing).

Issue #51, opened 2018-01-17
Use latest version of origin in tests


dmage:
We want to test the master code against the latest version of origin. See #50.

At the same time, the code from the release-3.8 branch is supposed to be tested with origin's branch release-3.8. And when we'll have the release-3.9 branch, it should use origin from the branch release-3.9.

bparees:
(we currently can't use "latest" because something changed in origin that is breaking our integration tests...sorting that out is the most urgent problem.  Getting our release branches using released versions of origin is secondary)

bparees:
(and getting to the point where things that break out integration tests can't merge into origin is also a high secondary priority)

stevekuznetsov:
You are OK building Origin in your jobs then, yeah? We'll just need to start the job from the base stage and build origin there/

bparees:
> You are OK building Origin in your jobs then, yeah? We'll just need to start the job from the base stage and build origin there/

yes. we need to build the right origin branch though.

stevekuznetsov:
Do you need intervention from me here? The branch of this repo being tested is provided as `${PULL_BASE_REF}` as specified [here](https://github.com/kubernetes/test-infra/tree/master/prow#job-evironment-variables). The step to build Origin would just be added in `aos-cd-jobs` as per usual.

PR #61, opened 2018-01-30
use versioned origin image for integration tests

bparees:
fixes https://github.com/openshift/image-registry/issues/51

### DIFF
--- a/pkg/testframework/versions.go
+++ b/pkg/testframework/versions.go
@@ -1,4 +1,17 @@
 package testframework
 
-// originImageRef is the Docker image name of the OpenShift Origin container.
-const originImageRef = "docker.io/openshift/origin:latest"
+import (
+	"os"
+)
+
+var (
+	// originImageRef is the Docker image name of the OpenShift Origin container.
+	originImageRef = "docker.io/openshift/origin:latest"
+)
+
+func init() {
+	version := os.Getenv("ORIGIN_VERSION")
+	if len(version) != 0 {
+		originImageRef = "docker.io/openshift/origin:" + version
+	}
+}


bparees:
@stevekuznetsov i'm interested in your thoughts on this and how it could/should be done differently to be more universally applicable across repositories/test jobs.

hack/test-integration.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/bash -e
+
+ORIGIN_VERSION=latest
+if [[ -n "$PULL_BASE_REF" && $PULL_BASE_REF != "master" ]]; then
```
stevekuznetsov: eek hate to leak ci into repos...

bparees: @stevekuznetsov sure. i'm open to suggestions......

stevekuznetsov: this logic in aos-cd-jobs?

stevekuznetsov: the script in the repo takes as input the version it needs?

pkg/testframework/versions.go
```diff
@@ -1,4 +1,17 @@
...
+	"os"
+)
+
+var (
+	// originImageRef is the Docker image name of the OpenShift Origin container.
+	originImageRef = "docker.io/openshift/origin:latest"
+)
+
+func init() {
+	version := os.Getenv("ORIGIN_VERSION")
```
bparees: this is now set via the job: https://github.com/openshift/aos-cd-jobs/pull/1053

bparees:
@dmage ptal

dmage:
@bparees do you want to merge it now or when the v3.9 tag will be available?

bparees:
@dmage let's hold it until the `origin:v3.9` tag that i've been promised, shows up.

@derekwaynecarr @smarterclayton any chance we can just manually create said tag now(based on what we currently call latest), since we're about to start cutting 3.9 branches and that would enable us to get our CI house in order?

/hold

bparees:
tags (e.g. origin:v3.9) have been created
/hold cancel

dmage:
[flake](https://openshift-gce-devel.appspot.com/build/origin-ci-test/pr-logs/pull/openshift_image-registry/61/test_pull_request_image_registry_extended/201/) openshift/origin#18384
/retest

PR #63, opened 2018-02-01
don't attempt manifest pullthrough against internal registry

bparees:
fixes https://github.com/openshift/origin/issues/18292


bparees:
@dmage ptal
/assign @dmage

bparees:
cloud provision

/retest

PR #65, opened 2018-02-02
Remove Tag/Untag

dmage:
Remove code that is not needed for our functionality.


pkg/dockerregistry/server/manifestservice.go
```diff
@@ -191,6 +191,15 @@ func (m *manifestService) Put(ctx context.Context, manifest distribution.Manifes
 // the content related to the manifest in the registry's storage (signatures).
 func (m *manifestService) Delete(ctx context.Context, dgst digest.Digest) error {
 	context.GetLogger(ctx).Debugf("(*manifestService).Delete")
+
+	_, _, err := m.imageStream.GetImageOfImageStream(ctx, dgst)
+	if err == nil {
```
bparees: this is slightly weird, can you add a comment explaining why we *expect* to get an error here?

dmage: Added, please check it.

dmage:
@bparees ImageStreamTags are created in the manifest service (using ImageStreamMappings). AFAIK, we do not expect the registry to remove tags.

This code is used, but it makes extra requests to do nothing: ManifestService should create manifests on the storage, TagService should tag them. We do not have such a separation, in our case the ManifestService creates a manifest and a tag, and TagService tags it once again.

We have this code just to implement the upstream interfaces.

pkg/dockerregistry/server/manifestservice.go
bparees: awesome, thanks

bparees:
Thanks for the summary, very helpful.  This lgtm but in deference to risk, let's wait until the release-3.9 branch is created (current target is wednesday) before merging this, so it gets a full release of testing instead of risking breaking 3.9 at the very end if some implication was overlooked.

pkg/dockerregistry/server/manifestservice.go
miminar: Why don't we delete the entry from imagestream?
We create the etcd entry on PUT. We should delete it on DELETE.

> The manifest is present *in the image stream*

Should be *in etcd*
(image stream doesn't contain it)

dmage: Do you want to remove the manifest from all tag's history?

We don't support PUT by digest, so it's ok to not implement DELETE by digest.

>> The manifest is present in the image stream
>
> Should be in etcd
> (image stream doesn't contain it)

You should read it as "At least one of the image stream's tags has in its history an item that references to an Image resource that has the same Name as the digest of the manifest."

miminar: > Do you want to remove the manifest from all tag's history?

Yes. I shouldn't be able to GET by digest from the repository after I DELETE by the digest.

> We don't support PUT by digest, so it's ok to not implement DELETE by digest.

At least we support PUT by tag, this PR removes all DELETE complements.

> You should read it as

I can translate the message but I doubt a newcomer can. I'd rather not comment at all than annotate with something untrue.

dmage: > Yes. I shouldn't be able to GET by digest from the repository after I DELETE by the digest.

Yes, I agree. But do you expect that it can affect other images in the repository? For example, that the entire tag with all its history is removed from the repository. Or that one of the tags is downgraded to the previous version of the image.

I don't expect neither of those. So for me the only solution is to forbid such deletions.

> At least we support PUT by tag, this PR removes all DELETE complements.

Yes, because the current implementation is broken.

> I can translate the message but I doubt a newcomer can. I'd rather not comment at all than annotate with something untrue.

This is not untrue.

miminar: > But do you expect that it can affect other images in the repository?

If you mean tags, then yes.

> For example, that the entire tag with all its history is removed from the repository. Or that one of the tags is downgraded to the previous version of the image.

As long as the history contains only the particular image, then yes to both. Upstream also deletes the relevant tags - all of them. As long as the user is authorized, I don't see a reason why we should prevent him from doing so.

> Yes, because the current implementation is broken.

You're right. Deleting entire tag with history regardless of other tagged images is wrong. But fixing it looks like a low hanging fruit to me, if we agree on a proper fix.

bparees:
/hold cancel

needs rebase though.

PR #67, opened 2018-02-02
bump(*)

bparees:
removes a couple unnecessary kube deps from glide.yaml


bparees:
@deads2k @dmage ptal

bparees:
/retest

Ben Parees | OpenShift

PR #73, opened 2018-03-06
Use SAR request to verify metrics access

legionus:
Cluster role: https://github.com/openshift/origin/pull/18968

### DIFF
--- a/pkg/dockerregistry/server/auth.go
+++ b/pkg/dockerregistry/server/auth.go
@@ -281,8 +281,8 @@ func (ac *AccessController) Authorized(ctx context.Context, accessRecords ...reg
 		case "metrics":
 			switch access.Action {
 			case "get":
-				if !isMetricsBearerToken(ac.metricsConfig, bearerToken) {
-					return nil, ac.wrapErr(ctx, ErrOpenShiftAccessDenied)
+				if err := verifyMetricsAccess(ctx, ac.metricsConfig, bearerToken, osClient); err != nil {
+					return nil, ac.wrapErr(ctx, err)
 				}
 			default:
 				return nil, ac.wrapErr(ctx, ErrUnsupportedAction)
@@ -402,21 +402,14 @@ func verifyWithSAR(ctx context.Context, resource, namespace, name, verb string, 
 	return nil
 }
 
-func verifyImageStreamAccess(ctx context.Context, namespace, imageRepo, verb string, c client.SelfSubjectAccessReviewsNamespacer) error {
-	return verifyWithSAR(ctx, "imagestreams/layers", namespace, imageRepo, verb, c)
-}
-
-func verifyImageSignatureAccess(ctx context.Context, namespace, imageRepo string, c client.SelfSubjectAccessReviewsNamespacer) error {
-	return verifyWithSAR(ctx, "imagesignatures", namespace, imageRepo, "create", c)
-}
-
-func verifyPruneAccess(ctx context.Context, c client.SelfSubjectAccessReviewsNamespacer) error {
+func verifyWithGlobalSAR(ctx context.Context, resource, subresource, verb string, c client.SelfSubjectAccessReviewsNamespacer) error {
 	sar := authorizationapi.SelfSubjectAccessReview{
 		Spec: authorizationapi.SelfSubjectAccessReviewSpec{
 			ResourceAttributes: &authorizationapi.ResourceAttributes{
-				Verb:     "delete",
-				Group:    imageapi.GroupName,
-				Resource: "images",
+				Verb:        verb,
+				Group:       imageapi.GroupName,
+				Resource:    resource,
+				Subresource: subresource,
 			},
 		},
 	}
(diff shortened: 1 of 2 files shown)


pkg/dockerregistry/server/auth.go
```diff
@@ -284,6 +284,9 @@ func (ac *AccessController) Authorized(ctx context.Context, accessRecords ...reg
 				if !isMetricsBearerToken(ac.metricsConfig, bearerToken) {
 					return nil, ac.wrapErr(ctx, ErrOpenShiftAccessDenied)
 				}
+				if err := verifyMetricsAccess(ctx, osClient); err != nil {
+					return nil, ac.wrapErr(ctx, err)
+				}
```
bparees: You need to combine this w/ the isMetricsBearerToken check...right now you're going to return an access denied before you even get here, right?

legionus: @bparees If the secret token is not true, then it makes no sense to continue. At the moment, a secret token is always required.

bparees: Hm, i would have expected us to be deprecating the secret token and not requiring it.

(Such that in 3.10 you can either supply the (deprecated) secret token, or a valid user auth header)

legionus: @bparees But I understand what you mean. It is necessary to support two options: with a secret and with a role. right ?

bparees: see above.  I'd expect us to deprecate the "with a secret" path.  What does the router offer?

legionus: @bparees You can set a static username/password or SAR check.

https://github.com/openshift/origin/blob/master/pkg/router/metrics/metrics.go#L52

bparees: Ok, I still think we should support "either/or" not "and" then.

legionus:
@bparees comments addressed. please review again.

bparees:
> Add cluster role like https://github.com/openshift/origin/blob/eaf3a0a732f8cf89606f7e7dfe3bf05b8a7b0a19/examples/prometheus/prometheus.yaml#L67-L77

or just add our resources to that role.  @smarterclayton @jwforres @jcantrill what is our plan around prometheus permissions?  Role per component, or a common "prometheus" role that can scrape all our infra components?

bparees:
/approve

lgtm but let's have @simo5 or @liggitt take a look since it's a security thing.

pkg/dockerregistry/server/auth.go
```diff
@@ -435,6 +427,37 @@ func verifyPruneAccess(ctx context.Context, c client.SelfSubjectAccessReviewsNam
...
+	}
+
+	if len(metrics.Secret) > 0 {
+		if metrics.Secret != token {
+			return ErrOpenShiftAccessDenied
+		}
+		return nil
+	}
+
+	if err := verifyWithGlobalSAR(ctx, "metrics", "get", c); err != nil {
```
liggitt: `metrics` aren't a resource in the `image.openshift.io` API group... might make more sense as a subresource (e.g. `images/metrics`) or as a nonResourceURL (e.g. `/metrics`)? @smarterclayton @sttts, do you know what are we checking for metrics permission for non-apiserver components?

legionus: @liggitt I assumed that it would be created role like:

https://github.com/openshift/origin/blob/8f44e6600770996be1acc29608b25e9d725d65eb/examples/prometheus/prometheus.yaml#L67-L77

openshift-docker:
Yeah, a virtual role and virtual permissions.

legionus:
Tested with the following cluster role:
```yaml
apiVersion: rbac.authorization.k8s.io/v1
kind: ClusterRole
metadata:
  name: prometheus-scraper
rules:
  - apiGroups:
    - image.openshift.io
    resources:
    - metrics
    verbs:
    - get
```

legionus:
[(5) Expose registry metrics with openshift auth [registry]](https://trello.com/c/iOKMS2PA/1490-5-expose-registry-metrics-with-openshift-auth-registry)

jcantrill:
@bparees  This is what we added in logging [1] with direction from @enj.  Which is guarded here [2].  Seems a fairly generically named role that could be reused.

[1] https://github.com/openshift/openshift-ansible/blob/master/roles/openshift_logging_elasticsearch/templates/2.x/logging-metrics-role.j2
[2] https://github.com/openshift/openshift-ansible/blob/master/roles/openshift_logging_elasticsearch/templates/2.x/es.j2#L145

enj:
Note that the linked `prometheus-metrics-viewer` is a `Role` (the intent being that it is not meant to be reused outside of the namespace).  If there is a use case for a role that needs to be used across multiple namespaces, it would make sense to switch the object to a `ClusterRole`.  That being said, I am not sure if the "metrics" being discussed in this PR are the same ones that `prometheus-metrics-viewer` is meant to guard.

bparees:
> That being said, I am not sure if the "metrics" being discussed in this PR are the same ones that prometheus-metrics-viewer is meant to guard.

@ironcladlou @liggitt @simo5 I think the three of you own the recommendation here:

1) what role should be defined for prometheus to use when scraping metrics from the registry?
2) will it be a global role used for all metrics scraping, or a role that specifically allows for registry metric scraping?

Landing this PR is effectively blocked until we have a decision on that so we can decide what the SAR check needs to look for.

simo5:
@bparees for (1) it depends on whether this action is part of a larger "role" prometheus assumes when performning this action such that you can use one single role to complete a task, or whether this si a very specific action that would make sense to (un)authorize individually
for (2) I have no input as I do not know what people are using this for, that should came from the PR author.

In general my advice is to think forward.
Is this functionality that is so core to prometheus that it has to be always granted ? If so then add this to the main prometheus role/rolebindings all in one place. If this is some optional activity or stuff we think may be disabled going forward or in specific environments then go with a spearate role or rolebinding that can be individually changed/deleted/protected etc...

bparees:
ok, from @smarterclayton on irc, we're supposed to update the existing prometheus cluster role:
https://github.com/openshift/origin/pull/18796/files#diff-d834a5fd4c91cb4a584e1113a1db37a5L67

to give that role permissions on a new virtual resource that we define, and then SAR check for that resource.

It seems like there's agreement we can make that virtual resource/subresource be "registry/metrics" and the apigroup would be "image.openshift.io"

@dmage please update this PR to add that permission to the prometheus role and change the SAR check to look for that.

jcantrill:
Or... could you reuse something like https://github.com/openshift/origin/pull/18796/files#diff-d834a5fd4c91cb4a584e1113a1db37a5R78

bparees:
> Or... could you reuse something like https://github.com/openshift/origin/pull/18796/files#diff-d834a5fd4c91cb4a584e1113a1db37a5R78

i can see that argument, but i also think it's reasonable to consider the image registry's metrics to be a part of the imageregistry api group, so i think we'll stick w/ that (it also has the benefit of having @smarterclayton's blessing and being aligned w/ the router did)

bparees:
sorry I mean to tag @legionus above, not @dmage

PR #79, opened 2018-05-03
Add pullthrough metrics

dmage:
https://trello.com/c/g3380oUO/1534-8-implement-recommended-registry-metrics-registry

New metrics:

  * imageregistry_pullthrough_blobstore_cache_requests_total{type}
  * imageregistry_pullthrough_repository_duration_seconds_bucket{registry,operation,le}
  * imageregistry_pullthrough_repository_duration_seconds_sum{registry,operation}
  * imageregistry_pullthrough_repository_duration_seconds_count{registry,operation}
  * imageregistry_pullthrough_repository_errors_total{registry,operation,code}

Cache request types:

  * Hit
  * Miss

Repository operations:

  * Init
  * BlobStore.Stat
  * BlobStore.Open
  * ManifestService.Get

PR #87, opened 2018-05-22
Registry catalog

miminar:
Implements the upstream repository catalog API call:

    GET /v2/_catalog

Returned is a list of repository names as json:

    200 OK
    Content-Type: application/json
    { "repositories": [ <name>, ... ] }

Pagination is supported by mapping the last returned repository name to an opaque continue token received from master API and using it on the next invocation.

Task: https://trello.com/c/AZINw5qI

Associated origin PR to allow registry to list image stream: openshift/origin#19879

Documentation: https://github.com/openshift/openshift-docs/pull/10102

**TODO:**
- [x] - tests
- [x] - use user token
- [x] - app holds just a lru cache

### DIFF
--- a/pkg/dockerregistry/server/app.go
+++ b/pkg/dockerregistry/server/app.go
@@ -8,6 +8,7 @@ import (
 	"github.com/docker/distribution/configuration"
 	"github.com/docker/distribution/context"
 	storagedriver "github.com/docker/distribution/registry/storage/driver"
+	kubecache "k8s.io/apimachinery/pkg/util/cache"
 
 	"github.com/openshift/image-registry/pkg/dockerregistry/server/cache"
 	"github.com/openshift/image-registry/pkg/dockerregistry/server/client"
@@ -21,6 +22,7 @@ const (
 	// Default values
 	defaultDescriptorCacheSize         = 4096
 	defaultDigestToRepositoryCacheSize = 2048
+	defaultPaginationCacheSize         = 1024
 )
 
 // appMiddleware should be used only in tests.
@@ -59,6 +61,10 @@ type App struct {
 
 	// metrics provide methods to collect statistics.
 	metrics metrics.Metrics
+
+	// paginationCache maps repository names to opaque continue tokens received from master API for subsequent
+	// list imagestreams requests
+	paginationCache *kubecache.LRUExpireCache
 }
 
 func (app *App) Storage(driver storagedriver.StorageDriver, options map[string]interface{}) (storagedriver.StorageDriver, error) {
@@ -66,9 +72,12 @@ func (app *App) Storage(driver storagedriver.StorageDriver, options map[string]i
 	return app.driver, nil
 }
 
-func (app *App) Registry(registry distribution.Namespace, options map[string]interface{}) (distribution.Namespace, error) {
-	app.registry = registry
-	return registry, nil
+func (app *App) Registry(nm distribution.Namespace, options map[string]interface{}) (distribution.Namespace, error) {
+	app.registry = nm
+	return &registry{
+		registry:   nm,
+		enumerator: NewCachingRepositoryEnumerator(app.registryClient, app.paginationCache),
+	}, nil
 }
 
 func (app *App) BlobStatter() distribution.BlobStatter {
@@ -82,11 +91,12 @@ func (app *App) BlobStatter() distribution.BlobStatter {
 // The program will be terminated if an error happens.
 func NewApp(ctx context.Context, registryClient client.RegistryClient, dockerConfig *configuration.Configuration, extraConfig *registryconfig.Configuration, writeLimiter maxconnections.Limiter) http.Handler {
 	app := &App{
-		ctx:            ctx,
-		registryClient: registryClient,
-		config:         extraConfig,
-		writeLimiter:   writeLimiter,
-		quotaEnforcing: newQuotaEnforcingConfig(ctx, extraConfig.Quota),
+		ctx:             ctx,
+		registryClient:  registryClient,
+		config:          extraConfig,
+		writeLimiter:    writeLimiter,
+		quotaEnforcing:  newQuotaEnforcingConfig(ctx, extraConfig.Quota),
+		paginationCache: kubecache.NewLRUExpireCache(defaultPaginationCacheSize),
 	}
 
 	if app.config.Metrics.Enabled {
--- a/pkg/dockerregistry/server/catalog.go
+++ b/pkg/dockerregistry/server/catalog.go
@@ -1,0 +1,163 @@
+package server
+
+import (
+	"errors"
+	"fmt"
+	"io"
+	"time"
+
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/util/cache"
+
+	"github.com/docker/distribution/context"
+
+	imageapiv1 "github.com/openshift/api/image/v1"
+	"github.com/openshift/image-registry/pkg/dockerregistry/server/client"
+)
+
+const paginationEntryTTL = time.Minute
+
+// RepositoryEnumerator allows to enumerate repositories known to the registry.
+type RepositoryEnumerator interface {
+	// EnumerateRepositories fills the given repos slice with image stream names. The slice's length
+	// determines the maximum number of repositories returned. The repositories are lexicographically sorted.
+	// The last argument allows for pagination. It is the offset in the catalog. Returned is a number of
+	// repositories filled. If there are no more repositories to return, io.EOF is returned.
+	EnumerateRepositories(ctx context.Context, repos []string, last string) (n int, err error)
+}
+
+// cachingRepositoryEnumerator is an enumerator that supports chunking by caching associations between
+// repository names and opaque continuation tokens.
+type cachingRepositoryEnumerator struct {
+	client client.RegistryClient
+	// a cache of opaque continue tokens for repository enumeration
+	cache *cache.LRUExpireCache
+}
+
+var _ RepositoryEnumerator = &cachingRepositoryEnumerator{}
+
+// NewCachingRepositoryEnumerator returns a new caching repository enumerator.
+func NewCachingRepositoryEnumerator(client client.RegistryClient, cache *cache.LRUExpireCache) RepositoryEnumerator {
+	return &cachingRepositoryEnumerator{
+		client: client,
+		cache:  cache,
+	}
+}
+
+type isHandlerFunc func(is *imageapiv1.ImageStream) error
+
+var errNoSpaceInSlice = errors.New("no space in slice")
+var errEnumerationFinished = errors.New("enumeration finished")
+
+func (re *cachingRepositoryEnumerator) EnumerateRepositories(
+	ctx context.Context,
+	repos []string,
+	last string,
+) (n int, err error) {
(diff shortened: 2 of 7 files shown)


pkg/dockerregistry/server/catalog.go
```diff
@@ -0,0 +1,136 @@
...
+
+func (r *cachingRepositoryEnumerator) enumerateImageStreams(
+	ctx context.Context,
+	limit int64,
+	last string,
+	handler isHandlerFunc,
+) error {
+	var start string
+
+	osClient, err := r.client.Client()
```
miminar: Probably user's token should be used instead. The registry cannot `list imagestreams` ATM, but we can be sure that user can thanks to the SAR check.

Otherwise `list imagestreams` needs to be added to `system:registry` cluster role.

pkg/dockerregistry/server/app.go
```diff
@@ -55,16 +57,44 @@ type App struct {
...
+
+func (r *registry) Scope() distribution.Scope {
+	return r.registry.Scope()
+}
+
+func (r *registry) Repository(ctx context.Context, name reference.Named) (distribution.Repository, error) {
+	return r.registry.Repository(ctx, name)
+}
+
+func (r *registry) Repositories(ctx context.Context, repos []string, last string) (n int, err error) {
```
miminar: Also a good candidate for metrics. Suitable as a follow-up after this and #79 are merged.

pkg/dockerregistry/server/app.go
```diff
@@ -55,16 +57,44 @@ type App struct {
 
 	// cache is a shared cache of digests and descriptors.
 	cache cache.DigestCache
+
+	// repositoryEnumerator allows to enumerate repository names known to registry
+	repositoryEnumerator RepositoryEnumerator
+}
+
+type registry struct {
```
miminar: Note to myself: Maybe worth its own file (to complement already existing `registry_test.go`).

pkg/dockerregistry/server/catalog.go
```diff
@@ -0,0 +1,136 @@
...
+
+var errEnumerationFinished = errors.New("enumeration finished")
+
+func (re *cachingRepositoryEnumerator) EnumerateRepositories(
+	ctx context.Context,
+	repos []string,
+	last string,
+) (n int, err error) {
+	if len(repos) == 0 {
+		// Client explicitely requested 0 results. Returning nil for error seems more appropriate but this is
```
bparees: explicitly

pkg/dockerregistry/server/catalog.go
```diff
@@ -0,0 +1,136 @@
...
+		// more in line with upstream does. Returning nil actually makes the upstream code panic.
+		// TODO: patch upstream?  /_catalog?n=0  results in 500
+		return 0, errors.New("no space in slice")
```
bparees: can we at least return a better error than "no space in slice", like "client requested zero entries"?

pkg/dockerregistry/server/catalog.go
bparees: using the user's token seems reasonable to me...

pkg/dockerregistry/server/catalog.go
dmage: If we'll use this API somewhere else, the message about the client will provide misleading information.

pkg/dockerregistry/server/catalog.go
```diff
@@ -0,0 +1,136 @@
...
+	}
+
+	warned := false
+	err = re.enumerateImageStreams(ctx, int64(len(repos)), last, func(is *imageapiv1.ImageStream) error {
+		name := fmt.Sprintf("%s/%s", is.Namespace, is.Name)
+		if len(last) > 0 && name <= last {
+			if !warned {
+				context.GetLogger(ctx).Warnf("pagination not working -> filtering the full image stream list")
```
dmage: What should I do when I see this warning in the log?

pkg/dockerregistry/server/catalog.go
```diff
@@ -0,0 +1,136 @@
...
+		return nil
+	})
+
+	switch err {
+	case errEnumerationFinished:
+		err = nil
+	case nil:
+		err = io.EOF
+	default:
+		context.GetLogger(ctx).Errorf("failed to enumerate repositories: %v", err)
```
dmage: Do we need this? I would expect the upstream code to log this error.

pkg/dockerregistry/server/catalog.go
miminar: Check if you run the registry against different version of master API. The message seemed too long already but I can add the suggestion.

dmage: Can this situation be handled by enumerateImageStreams, so that we don't need to spread this hack by the code?

pkg/dockerregistry/server/catalog.go
```diff
@@ -0,0 +1,136 @@
...
+		} else {
+			start = c.(string)
+		}
+	}
+	iss, err := osClient.ImageStreams("").List(metav1.ListOptions{
+		Limit:    limit,
+		Continue: start,
+	})
+
+	if err != nil {
```
dmage: The master may respond with 410 ResourceExpired, that should be treated as an indication to repeat the query without Continue.

pkg/dockerregistry/server/catalog.go
```diff
@@ -0,0 +1,136 @@
...
+) error {
+	var start string
+
+	osClient, err := r.client.Client()
+	if err != nil {
+		return err
+	}
+
+	if len(last) > 0 {
+		if c, ok := r.cache.Get(last); !ok {
```
dmage: Can we use this token twice?

pkg/dockerregistry/server/catalog.go
miminar: Good point, I'll handle the case.

pkg/dockerregistry/server/catalog.go
miminar: Yes, from what I understand, the same token can be returned to two different clients or to two different requests.

pkg/dockerregistry/server/catalog.go
miminar: Sure. I'll rework it.

pkg/dockerregistry/server/catalog.go
bparees: still seems like there is a more informative message we can provide/bubble up.  it's pretty much a self-inflicted+correctable error, so the error text should make that clear..  "no space in slice" makes it sound like an internal error where we ran out of memory and nothing can be done w/o changing code.

miminar:
Addressed most of the comments.

pkg/dockerregistry/server/catalog.go
```diff
@@ -0,0 +1,164 @@
...
+		osClient, err := r.client.Client()
+		if err != nil {
+			return err
+		}
+		client = osClient
+	}
+
+	if len(last) > 0 {
+		if c, ok := r.cache.Get(last); !ok {
+			context.GetLogger(ctx).Warnf("failed to find opaque continue token for last repository=%q -> requesting the full image stream list instead of %d items", last, limit)
```
bparees: should set warned=true?

miminar: Not necessary, but it will add to legibility and robustness.

pkg/dockerregistry/server/catalog.go
miminar: You are right, the returned error is logged, I've removed the logging statement.

miminar:
Added unit tests  and squashed.

pkg/dockerregistry/server/catalog.go
miminar: Added

dmage:
My main concern that it will perform poorly behind load balancers without sticky sessions. And as I can see, the optimistic way when we have an opaque token is not tested.

If we are fine with slow requests, then LGTM.

bparees:
> My main concern that it will perform poorly behind load balancers without sticky sessions.

specifically the pagination you mean?

miminar:
Our docker-registry's service has `sessionAffinity=ClientIP` by default. We can only hope that clients will not be too dumm to ask for all.

miminar:
@bmcelvee PTAL. No rush, waits for the blocked PRs.

miminar:
Rebased.

miminar:
Re: `ci/prow/e2e - Job failed.` - Looks like broken environment:

> failed to build "172.30.42.94:5000/e2e-test-prune-images-hnpr6/prune-schema2-all-images-false:latest" image: dial unix /var/run/docker.sock: connect: no such file or directory

/retest

miminar:
@openshift/sig-continuous-infrastructure to rescue:

> failed to build "172.30.74.83:5000/e2e-test-prune-images-dt99r/prune:latest" image: dial unix /var/run/docker.sock: connect: no such file or directory

smarterclayton:
Does test-prune-images assume it has access to the docker socket on the host where the e2e test runs?  IF so it should be `[local]` and excluded, or fixed to not assume that.

bparees:
yes it looks like hardprune.go is building an image locally to push into the registry in order to tightly control the structure of the image being produced.

bparees:
@smarterclayton if we mark it [local], under what conditions will it still get run?

bparees:
(marking it local here: https://github.com/openshift/origin/pull/20061)

openshift-docker:
When run as part of a single node installation.

Extended tests shouldn’t depend on local access to the node.  Use a
privileged pod if you need to check host stuff (and remember, you would
fail on crio).

On Jun 21, 2018, at 1:11 PM, Ben Parees <notifications@github.com> wrote:

@smarterclayton <https://github.com/smarterclayton> if we mark it [local],
under what conditions will it still get run?

—
You are receiving this because you are subscribed to this thread.
Reply to this email directly, view it on GitHub
<https://github.com/openshift/image-registry/pull/87#issuecomment-399177226>,
or mute the thread

bparees:
> When run as part of a single node installation.

@smarterclayton right, so does any of our PR testing on either origin or image-registry exercise the extended tests on a single node installation?  That is my fundamental question here...  once i tag the test as  `[local]` will we still know if something breaks it.

> Extended tests shouldn’t depend on local access to the node.  Use a
privileged pod if you need to check host stuff (and remember, you would
fail on crio).

I get it, trying to evaluate how urgent it is that we fix this so the test is being executed again.

smarterclayton:
On Jun 21, 2018, at 6:36 PM, Ben Parees <notifications@github.com> wrote:

When run as part of a single node installation.

@smarterclayton <https://github.com/smarterclayton> right, so does any of
our PR testing on either origin or image-registry exercise the extended
tests on a single node installation? That is my fundamental question
here... once i tag the test as [local] will we still know if something
breaks it.



The master AWS job currently tests it.  We will probably have a single node
oc cluster up test that may be able to use this in the future.


Extended tests shouldn’t depend on local access to the node. Use a
privileged pod if you need to check host stuff (and remember, you would
fail on crio).

I get it, trying to evaluate how urgent it is that we fix this so the test
is being executed again.

—
You are receiving this because you were mentioned.
Reply to this email directly, view it on GitHub
<https://github.com/openshift/image-registry/pull/87#issuecomment-399264597>,
or mute the thread

bparees:
> The master AWS job currently tests it.  We will probably have a single node
oc cluster up test that may be able to use this in the future.

ok, good enough, thanks.

miminar:
Rebased.

PR #88, opened 2018-05-23
Support TEST_KUBECONFIG for integration tests

smarterclayton:
Uses the provided administrator kubeconfig and runs the tests against
that master. This allows the integration tests to be run against a stock
"openshift start master".

Another change in the master will make local dev cases easier so allowed
registries doesn't have to be customized.

@bparees @legionus

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -105,7 +105,7 @@ test-unit:
 
 test-integration:
 	# TODO(dmage): remove DOCKER_API_VERSION when our CI will upgrade Docker
-	DOCKER_API_VERSION=1.24 go test ./test/integration/...
+	GOTEST_FLAGS="$(TESTFLAGS)" DOCKER_API_VERSION=1.24 hack/test-go.sh test/integration/*
 .PHONY: test-integration
 
 # Remove all build artifacts.
--- a/pkg/testframework/master.go
+++ b/pkg/testframework/master.go
@@ -11,6 +11,7 @@ import (
 	"net/http"
 	"os"
 	"path"
+	"path/filepath"
 	"strconv"
 	"testing"
 	"time"
@@ -22,6 +23,7 @@ import (
 	"github.com/docker/docker/pkg/archive"
 
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/apimachinery/pkg/util/wait"
 	kubeclient "k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/rest"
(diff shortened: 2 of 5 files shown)


smarterclayton:
Here's what the runner will look like:

```
    # Once admin.kubeconfig exists, executes shared tests
    - name: test
      image: ${LOCAL_IMAGE_TEST_BIN}
      volumeMounts:
      - name: shared-tmp
        mountPath: /tmp/shared
      - name: artifacts
        mountPath: /tmp/artifacts
      command:
      - /bin/bash
      - -c
      - |
        #!/bin/bash
        set -euo pipefail

        trap 'touch /tmp/shared/exit' EXIT
        trap 'kill $(jobs -p); exit 0' TERM

        # wait until the master job creates admin.kubeconfig
        while true; do
          if [[ ! -f /tmp/shared/admin.kubeconfig ]]; then
            sleep 5 & wait
            continue
          fi
          break
        done
        echo "Found kubeconfig"
        export KUBECONFIG=/tmp/shared/admin.kubeconfig
        
        ${COMMAND}

    # Start a standalone master
    - name: control-plane
      image: ${IMAGE_CONTROL_PLANE}
      volumeMounts:
      - name: shared-tmp
        mountPath: /tmp
      - name: artifacts
        mountPath: /tmp/artifacts
      env:
      - name: TYPE
        value: ${CLUSTER_TYPE}
      command:
      - /bin/bash
      - -c
      - |
        #!/bin/bash
        set -euo pipefail
        function teardown() {
          set +e
          echo "Gathering artifacts ..."
        }
        trap 'teardown' EXIT
        trap 'kill $(jobs -p); exit 0' TERM

        ( 
          while true; do
            if [[ -f /tmp/exit ]]; then
              echo "Shutting down"
              kill $$
            fi
            sleep 5 & wait
          done
        ) &

        mkdir /tmp/cluster && cd /tmp/cluster
        openshift start master --write-config=/tmp
        openshift start master --config=/tmp/master-config.yaml
```

smarterclayton:
That runner would be shared between test-cmd and this integration test and anyone else who just needs a stock origin master.  When we split origin master up, we'll update the shared runner so you don't have to worry about it.

smarterclayton:
Linked the change to openshift to stop defaulting whitelisted registries, to be consistent with install.

smarterclayton:
I linked in the integration job - if we're ok with these changes, then I'll test those and iterate, but integration should take ~10s with master reuse vs 1m or so.

bparees:
so this is just establishing an abstraction where you can either run a master in a container (old behavior) or point the client to an existing cluster via an admin kubeconfig (new behavior), right?

looks reasonable to me as long as the old behavior wasn't customizing the master it started in any meaningful way.

smarterclayton:
Once this is merged I'll start testing the new job against my extra PR.

smarterclayton:
Testing the job now

smarterclayton:
Now waiting on https://github.com/openshift/origin/pull/19829

openshift-docker:
e2e is broken because of an accident with azure, fixing at the source.

PR #86, opened 2018-05-22
Provide clients with additional information on failures

dmage:
Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1578269.
<details>
<summary>Output of `docker pull 172.30.1.1:5000/myproject/foo` before patch</summary>

> Using default tag: latest
[...]

PR #93, opened 2018-06-04
Remove remote stat from blobDescriptorService

dmage:


bparees:
@dmage what made this code unnecessary?

legionus:
> what made this code unnecessary?

We verify the blobs/manifests in our code. We can't do it in upstream any more since we are doing more than them. The `BlobDescriptorService`  should no longer be called for non-local blobs.

dmage:
@bparees this code was added (as I can see) to give the upstream code knowledge about pullthrough. But later we implemented verification on our side and now it works with our pullthroughBlobStore. The upstream code isn't involved in verification anymore, thus this code can be removed.

It's a part of the long running story about untangling our code. pkg/dockerregistry/server contains a lot of code which is hard to decouple. It makes code harder to test and debug. As a visible consequence: it prevents me from doing #86. Also it complicates fixing our logging.

PR #96, opened 2018-06-06
remove unused counter variable

bparees:


bparees:
@legionus @dmage ptal.  seems like this counter isn't being used for anything anyway?

PR #101, opened 2018-06-12
Perform an O(1) check to find layers in an image

smarterclayton:
Using the new imagestream/layers API, avoid making O(N) image fetches in the case where we don't have a cached association for the reference.

API in https://github.com/openshift/origin/pull/19969

### DIFF
--- a/hack/test-local-registry.sh
+++ b/hack/test-local-registry.sh
@@ -1,0 +1,26 @@
+#!/bin/bash
+
+#
+# This script launches the image registry pointing to the OpenShift master defined in
+# the KUBECONFIG environment variable (and uses the current credentials as the client
+# credentials). You may need to start your master with the internalRegistryHostname
+# configuration variable set for some image stream calls to function properly.
+#
+
+source "$(dirname "${BASH_SOURCE}")/lib/init.sh"
+
+os::build::setup_env
+
+os::util::ensure::built_binary_exists 'dockerregistry'
+
+url="${REGISTRY_OPENSHIFT_SERVER_ADDR:-localhost:5000}"
+# find the first builder service account token
+token="$(oc get $(oc get secrets -o name | grep builder-token | head -n 1) --template '{{ .data.token }}' | os::util::base64decode)"
+echo
+echo "Login with:"
+echo "  docker login -p \"${token}\" -u user ${url}"
+echo
+
+REGISTRY_STORAGE_FILESYSTEM_ROOTDIRECTORY="${REGISTRY_STORAGE_FILESYSTEM_ROOTDIRECTORY:-/tmp/registry}" \
+  REGISTRY_OPENSHIFT_SERVER_ADDR="${url}" \
+	dockerregistry images/dockerregistry/config.yml
--- a/pkg/dockerregistry/server/blobdescriptorservice.go
+++ b/pkg/dockerregistry/server/blobdescriptorservice.go
@@ -40,28 +40,32 @@ func (bs *blobDescriptorService) Stat(ctx context.Context, dgst digest.Digest) (
 	if err == nil {
 		return desc, nil
 	}
-
 	context.GetLogger(ctx).Debugf("(*blobDescriptorService).Stat: could not stat layer link %s in repository %s: %v", dgst.String(), bs.repo.Named().Name(), err)
 
+	// we couldn't find the layer link
 	desc, err = bs.repo.app.BlobStatter().Stat(ctx, dgst)
-	if err == nil {
-		context.GetLogger(ctx).Debugf("(*blobDescriptorService).Stat: blob %s exists in the global blob store", dgst.String())
-		// only non-empty layers is wise to check for existence in the image stream.
-		// schema v2 has no empty layers.
-		if !isEmptyDigest(dgst) {
-			// ensure it's referenced inside of corresponding image stream
-			if bs.repo.cache.ContainsRepository(dgst, bs.repo.imageStream.Reference()) {
-				context.GetLogger(ctx).Debugf("found cached blob %q in repository %s", dgst.String(), bs.repo.imageStream.Reference())
-			} else if image := bs.repo.imageStream.HasBlob(ctx, dgst); image != nil {
-				// remember all the layers of matching image
-				RememberLayersOfImage(ctx, bs.repo.cache, image, bs.repo.imageStream.Reference())
-			} else {
-				context.GetLogger(ctx).Debugf("(*blobDescriptorService).Stat: blob %s is neither empty nor referenced in image stream %s", dgst.String(), bs.repo.Named().Name())
-				return distribution.Descriptor{}, distribution.ErrBlobUnknown
-			}
-		}
+	if err != nil {
+		return desc, err
+	}
+	context.GetLogger(ctx).Debugf("(*blobDescriptorService).Stat: blob %s exists in the global blob store", dgst.String())
+
+	// Empty layers are considered to be "public" and we don't need to check whether they are referenced - schema v2
+	// has no empty layers.
+	if isEmptyDigest(dgst) {
 		return desc, nil
 	}
 
-	return desc, err
+	// ensure it's referenced inside of corresponding image stream
+	if bs.repo.cache.ContainsRepository(dgst, bs.repo.imageStream.Reference()) {
+		context.GetLogger(ctx).Debugf("(*blobDescriptorService).Stat: found cached blob %q in repository %s", dgst.String(), bs.repo.imageStream.Reference())
+		return desc, nil
+	}
+	if image := bs.repo.imageStream.HasBlob(ctx, dgst); image != nil {
+		// remember all the layers of matching image
+		RememberLayersOfImage(ctx, bs.repo.cache, image, bs.repo.imageStream.Reference())
+		return desc, nil
+	}
+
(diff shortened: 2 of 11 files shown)


smarterclayton:
@bparees

hack/test-local-registry.sh
```diff
@@ -0,0 +1,18 @@
...
+# find the first builder service account token
+token="$(oc get $(oc get secrets -o name | grep builder-token | head -n 1) --template '{{ .data.token }}' | os::util::base64decode)"
+echo
+echo "Login with:"
+echo "  docker login -p \"${token}\" -u user ${url}"
+echo
+
+REGISTRY_STORAGE_FILESYSTEM_ROOTDIRECTORY="${REGISTRY_STORAGE_FILESYSTEM_ROOTDIRECTORY:-/tmp/registry}" \
+  REGISTRY_OPENSHIFT_SERVER_ADDR="${url}" \
+	dockerregistry images/dockerregistry/config.yml
```
bparees: i assume this expects a cluster to be running somewhere, how are the coordinates of the cluster being supplied to the registry?

pkg/dockerregistry/server/blobdescriptorservice.go
```diff
@@ -40,28 +40,32 @@ func (bs *blobDescriptorService) Stat(ctx context.Context, dgst digest.Digest) (
 	if err == nil {
 		return desc, nil
 	}
-
 	context.GetLogger(ctx).Debugf("(*blobDescriptorService).Stat: could not stat layer link %s in repository %s: %v", dgst.String(), bs.repo.Named().Name(), err)
 
+	// we couldn't find the layer link
```
bparees: comment should elucidate the implications of not finding the layer link

pkg/imagestream/imagestreamhasblob.go
```diff
@@ -37,6 +38,27 @@ func (is *imageStream) HasBlob(ctx context.Context, dgst digest.Digest, requireM
...
+		copied := make([]imageapiv1.ImageLayer, 0, len(layers.Blobs))
+		for name, blob := range layers.Blobs {
+			copied = append(copied, imageapiv1.ImageLayer{Name: name, LayerSize: blob.LayerSize, MediaType: blob.MediaType})
+		}
+		var firstImage string
+		for name := range layers.Layers {
+			firstImage = name
+			break
+		}
+		return logFound(&imageapiv1.Image{
```
bparees: what is layers.Layers was empty?

pkg/imagestream/imagestreamhasblob.go
```diff
@@ -37,6 +38,27 @@ func (is *imageStream) HasBlob(ctx context.Context, dgst digest.Digest, requireM
...
+		if _, ok := layers.Blobs[dgst.String()]; !ok {
+			return logFound(nil)
+		}
+		copied := make([]imageapiv1.ImageLayer, 0, len(layers.Blobs))
+		for name, blob := range layers.Blobs {
+			copied = append(copied, imageapiv1.ImageLayer{Name: name, LayerSize: blob.LayerSize, MediaType: blob.MediaType})
+		}
+		var firstImage string
+		for name := range layers.Layers {
+			firstImage = name
```
bparees: we don't actually need to find the specific image that contains the blob?  any image in the imagestream will do?

vendor/github.com/openshift/client-go/image/clientset/versioned/typed/image/v1/imagestream_expansion.go
```diff
@@ -0,0 +1,39 @@
...
+import (
+	v1 "github.com/openshift/api/image/v1"
+)
+
+// The ImageStreamExpansion interface allows manually adding extra methods to the ImageStream interface.
+type ImageStreamExpansion interface {
+	Layers(name string) (*v1.ImageStreamLayers, error)
+}
+
+// Bind applies the provided binding to the named pod in the current namespace (binding.Namespace is ignored).
```
bparees: godoc is wrong

hack/test-local-registry.sh
smarterclayton: KUBECONFIG

pkg/dockerregistry/server/blobdescriptorservice.go
smarterclayton: this is just a refactor of the method, i'm happy to make better comments

pkg/imagestream/imagestreamhasblob.go
smarterclayton: that means the image stream has no layers

pkg/imagestream/imagestreamhasblob.go
smarterclayton: correct

pkg/imagestream/imagestreamhasblob.go
bparees: yeah but we shouldn't returning found in that case, right?

smarterclayton: if we don't have a layer we return on line 43 with nil, and the if block fails, and we return ErrBlobUnknown.

hack/test-local-registry.sh
bparees: add a usage comment at the top of the file?

pkg/imagestream/imagestreamhasblob.go
```diff
@@ -37,6 +38,27 @@ func (is *imageStream) HasBlob(ctx context.Context, dgst digest.Digest, requireM
...
+			break
+		}
+		return logFound(&imageapiv1.Image{
+			ObjectMeta: metav1.ObjectMeta{
+				Name: firstImage,
+			},
+			DockerImageLayers: copied,
```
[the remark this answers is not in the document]
dmage: RememberLayersOfImage will report the error `"image %s does not have metadata"` for this image. We trying to reduce amount of logs.

pkg/imagestream/imagestreamhasblob.go
legionus: Regression. The `requireManaged` was ignored. It will break `config.Pullthrough.Enabled` flag.

https://github.com/openshift/image-registry/pull/101/files#diff-0c454c8033229f303d88c165f82eb0d9L82

smarterclayton: I’m not sure pullthrough enabled false is something we should support.  In 8 release we’ve never had a reason to trn it off.  At this point, I’d probably prefer to drop the use case.

I don’t really consider the distinction between managed an unmanaged images important anymore.  If a layer is available it should be served.  The image it comes from is irrelevant.

pkg/imagestream/imagestreamhasblob.go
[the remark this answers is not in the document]
smarterclayton: I’m going to remove that message.  This is purely a blob access check.

smarterclayton:
I still need to add the e2e test to try and access content.  Will do that once we clear up the case.

pkg/imagestream/imagestreamhasblob.go
smarterclayton: Specifically, for the case where we are attempting to find the layers, I think this method should be returning a set of layers that are considered valid (which is added to the cache).

legionus: OK. Please, open an issue to make pullthrough not optional to address this important change. Because right now this functionality can be turned off and removing this option will need to be written to the log as a warning in the case of `config.Pullthrough.Enabled = false`.

/cc @bparees @dmage

pkg/imagestream/imagestreamhasblob.go
dmage: If we have the list of blobs, we don't need the image object anymore, we should just return that list.

pkg/imagestream/imagestreamhasblob.go
```diff
@@ -37,6 +38,27 @@ func (is *imageStream) HasBlob(ctx context.Context, dgst digest.Digest, requireM
 		return found
 	}
 
+	if layers, err := is.imageStreamGetter.layers(); err == nil {
```
dmage: Why do we need to fallback to the old code?

smarterclayton: For 1 release we should tolerate the cluster not having /layers or the admin not configuring permissions.  We can then remove the old code path in 3.12.

pkg/imagestream/imagestreamhasblob.go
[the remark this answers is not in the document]
smarterclayton: Then I'll split this into RememberLayers and RememberLayersOfImage.

smarterclayton:
Updated with the latest version of the api, all feedback added to the PR.

The biggest change is the signature to `HasBlob` - it now returns `(bool, *ImageStreamLayers, *Image)`, and the caller uses the two objects if they are available as "things to add to the cache".  When we remove the legacy path in 3.12/4.0 we can drop that segment and simplify the signature back down.

Unit tests added, the e2e test will be added to the origin PR.

bparees:
looks reasonable to me

smarterclayton:
Bumped dependencies in #113, which I will rebase on master after that merges.

hack/test-local-registry.sh
smarterclayton: Done

smarterclayton:
Updated on bumps, ready for final review.

Added one commit to switch us to building on go1.10 for travis.

bparees:
lgtm.  Any final concerns @dmage or @legionus ?

smarterclayton:
No comments?  I'm happy to do follow ups (I need to improve the e2e tests in origin for pulling images.

PR #103, opened 2018-06-14
pullthrough is not optional

legionus:
> I’m not sure pullthrough enabled false is something we should support. In 8 release we’ve never had a reason to trn it off. (c) Clayton Coleman

https://github.com/openshift/origin/pull/19969#discussion_r195331219
https://github.com/openshift/image-registry/pull/101#discussion_r195322187

/cc @smarterclayton

### DIFF
--- a/pkg/dockerregistry/server/registry_test.go
+++ b/pkg/dockerregistry/server/registry_test.go
@@ -37,7 +37,6 @@ func newTestRegistry(
 	osClient registryclient.Interface,
 	storageDriver driver.StorageDriver,
 	blobrepositorycachettl time.Duration,
-	pullthrough bool,
 	useBlobDescriptorCacheProvider bool,
 ) (distribution.Namespace, error) {
 	cfg := &configuration.Configuration{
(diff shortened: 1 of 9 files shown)


pkg/dockerregistry/server/registry_test.go
```diff
@@ -36,15 +36,14 @@ func newTestRegistry(
...
-	pullthrough bool,
 	useBlobDescriptorCacheProvider bool,
 ) (distribution.Namespace, error) {
 	cfg := &configuration.Configuration{
 		Server: &configuration.Server{
 			Addr: "localhost:5000",
 		},
 		Pullthrough: &configuration.Pullthrough{
-			Enabled: pullthrough,
+			Enabled: true,
```
dmage: Can we just drop this parameter from the configuration?

legionus: @dmage I want to show warning if someone have it as `false`. But I can drop the `Pullthrough` initialization here.

smarterclayton:
Ben and I agree - we have no reason to keep supporting this.   Approval
granted to terminate this code path.

legionus:
Rebased

PR #105, opened 2018-06-22
Return 404 when image stream doesn't exist

dmage:
https://bugzilla.redhat.com/show_bug.cgi?id=1593795


bparees:
@dmage can you make a PR against the 3.10 branch too?  @smarterclayton wants this in 3.10.

PR #108, opened 2018-06-29
switch glide to openshift release-3.10 branches

bparees:


bparees:
@dmage ok 3.10 has shipped so this should be safe to merge.  ptal.

PR #115, opened 2018-07-28
Treat manifests when blobs when deciding to serve them

smarterclayton:
Manifests are content addressible and are valid blob retrieval targets.


smarterclayton:
Discovered this when updating api.ci.  The e2e test we need is a cross repo tag and mount, then attempting to access the blobs for the image.  The layers e2e test in origin is a start, I need to add this scenario in.

Also discovered that the role was missing for system:registry (will fix in a separate PR).

pkg/dockerregistry/server/util.go
```diff
@@ -84,4 +84,7 @@ func RememberLayersOfImageStream(ctx context.Context, cache cache.RepositoryDige
 	for dgst := range layers.Blobs {
 		_ = cache.AddDigest(digest.Digest(dgst), cacheName)
 	}
+	for dgst := range layers.Images {
+		_ = cache.AddDigest(digest.Digest(dgst), cacheName)
+	}
```
bparees: dumb question but since the manifest is just another blob, why isn't it part of the layers.Blobs iteration already?

smarterclayton: Actually.... that's a good question.

smarterclayton: You're right, also, the API for ImageLayers is wrong.  Opening an origin PR to resolve the internal problem and fix the code issue, then I'll change the external API to clarify.

Once that's done we can then close this.

smarterclayton: Moved to https://github.com/openshift/origin/pull/20458.  That's the first fix.

Will change the public API in a subsequent PR.

dmage:
> Manifests are content addressible and are valid blob retrieval targets.

Why do you think so? AFAIK this is an incorrect statement.

```
$ boater get-manifest ubuntu@sha256:3f119dc0737f57f704ebecac8a6d8477b0f6ca1ca0332c7ee1395ed2c6a82be7 | head -n3
{
   "schemaVersion": 2,
   "mediaType": "application/vnd.docker.distribution.manifest.list.v2+json",
$ boater get-blob ubuntu sha256:3f119dc0737f57f704ebecac8a6d8477b0f6ca1ca0332c7ee1395ed2c6a82be7
{"errors":[{"code":"BLOB_UNKNOWN","message":"blob unknown to registry","detail":"sha256:3f119dc0737f57f704ebecac8a6d8477b0f6ca1ca0332c7ee1395ed2c6a82be7"}]}
```

openshift-docker:
Him - the original error was because during docker pull we were failing in
blob descriptor stat HasBlob to access the image digest (not the config
blob).  Did we start calling blob descriptor service on a manifest (and we
shouldn't be)?

smarterclayton:
I think this is because BlobAccessController is calling the API, and blob access controller is called for manifests.  Is that not the case?

smarterclayton:
Looking at manifest store, it certainly looks like it's accessing the blob store for that content.  I don't see how the blob store filters out manifests?

smarterclayton:
If the internal registry is asking for blobs via a blob access controller and that blob is a manifest (and stored underneath the covers), then I'd probably say for the intent of the layer API the manifest *is* a blob.  It's just not a layer or a config blob.

bparees:
>  I'd probably say for the intent of the layer API the manifest is a blob. It's just not a layer or a config blob.

makes sense to me (i'd expect config blobs to work that way also).

dmage:
https://github.com/docker/distribution/blob/0dae0957e5fe3156c265d22bef4cba9efbf388e2/registry/storage/registry.go#L220
https://github.com/docker/distribution/blob/0dae0957e5fe3156c265d22bef4cba9efbf388e2/registry/storage/registry.go#L287

It's the same linked blob store with different path specs. In upstream it works because BlobAccessController uses different statters.

However, the same BlobAccessController is used for both stores, and it ignores results from the upstream statter. It means we have been considering manifests as regular blobs for a long time.

Though, it's a bug on the registry side and one day we may want to fix it, therefore it's better to keep manifests and blobs separate in the layers API.

smarterclayton:
Ok, so then after the API change goes in (to rename the manifest -> config, which was always wrong), this PR will consider the image as a serving and I'll add a comment.  Let's get final review on the API change.

bparees:
> this PR will consider the image as a serving

?

smarterclayton:
The PR will change so that if it sees an image in the image layers object it considers that name acceptable for blob serving (serves the manifest to preserve backwards compatibility with our previous registry versions)

smarterclayton:
Please take a look, the api change is pending.

Serves a blob if it’s digest matches a layer, the config json, or the image digest (for the manifest).

bparees:
> Please take a look, the api change is pending.

which api change/where?

otherwise lgtm.

smarterclayton:
API changes are in to origin - any other comments?

bparees:
/lgtm

unhold at your discretion.

PR #116, opened 2018-08-01
Rebase

dmage:


bparees:
/hold

We're going to want to hold this until we branch for 3.11 (we don't want to do this in 3.11, due to the risk).

dmage:
@bparees @legionus https://github.com/openshift/docker-distribution/commit/0613ce94029ba3aba1a9db12d3b4fdc4cf0d5a98 - do we need this carry patch?

I can't imagine a situation when we need to pull from a remote registry by tag, usually we get digests from the imagestream API.

bparees:
> @bparees @legionus openshift/docker-distribution@0613ce9 - do we need this carry patch?

i'm not even sure we need it in origin anymore?  Doesn't our image import logic support manifestlists now (by converting them/extracting a single architecture manifest from the manifestlist)?

dmage:
the problem I've found is kind of fixed

bparees:
> the problem I've found is kind of fixed

@dmage fixed to the point where we want to merge it?

dmage:
@bparees yes, we can merge it. There was a problem with contexts (the request context isn't inherited from the application context anymore) and, as I can see, it affected only logging in the tests. It was fixed by a carry patch.

bparees:
hooray!

PR #118, opened 2018-08-08
Fix waiting for cluster roles

dmage:
An attempt to address https://github.com/openshift/test-infra/issues/164.

### DIFF
--- a/pkg/testframework/master.go
+++ b/pkg/testframework/master.go
@@ -22,6 +22,7 @@ import (
 	"github.com/docker/docker/client"
 	"github.com/docker/docker/pkg/archive"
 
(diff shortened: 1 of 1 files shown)


dmage:
CI flake
/retest

dmage:
```
got error on docker inspect fork/exec /bin/docker: no such file or directory
```
/retest

dmage:
oh c'mon
/retest

bparees:
there was some chatter on aos-devel about the merge queue being disabled due to github apis being broken.  I don't think it would cause failing tests, but fyi this may not be mergeable until that is resolved anyway.

bparees:
(though we can also hand merge it if the tests are all passing)

PR #123, opened 2018-09-19
Fix /v2/ endpoint

dmage:
Fixes [bug 1630537](https://bugzilla.redhat.com/show_bug.cgi?id=1630537).


bparees:
We need to understand why that bug hit this, because that bug was filed against a v3.11 PR, it should not be testing origin master code.

Furthermore, the bug was not manifesting in origin master PRs.

/lgtm

dmage:
@bparees origin's end_to_end test is the same for all branches:

https://github.com/openshift/release/blob/13517c7eb2290c381124e108f026a55f7a3325ae/ci-operator/jobs/openshift/origin/openshift-origin-presubmits.yaml#L402-L409

dmage:
3.10 is affected as well: https://github.com/openshift/origin/pull/21013

PR #125, opened 2018-09-21
Update deps

### LINKED ISSUE #112: image-registry is tied to Sirupsen (uppercase)
This means we can't pick up newer kube deps (which are tied to sirupsen lowercase)

It looks like the master branch of docker distribution has handled this(moved to lowercase), so we should probably do what we proposed a few weeks ago and move up to master.

The only alternative I see (other than getting left behind on really old kube deps) would be to do a carry patch for docker distribution to switch it to lowercase sirupsen and that seems terrible.

@legionus @dmage any other options you see?

legionus:
Fixes #112


dmage:
@legionus the test for `sha256:foo` needs to be updated, everything else looks good to merge.

legionus:
> the test for sha256:foo needs to be updated, everything else looks good to merge.

@dmage yep. the error has  changed. I already found it. Fixed.

bparees:
thanks @legionus !

PR #128, opened 2018-10-10
Require Authorization header only if we do not have permissions for pull/push

legionus:
Please review it very carefully.

Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1636209

### DIFF
--- a/pkg/dockerregistry/server/auth.go
+++ b/pkg/dockerregistry/server/auth.go
@@ -187,7 +187,9 @@ func (ac *AccessController) Authorized(ctx context.Context, accessRecords ...reg
 
 	bearerToken, err := getOpenShiftAPIToken(req)
 	if err != nil {
-		return nil, ac.wrapErr(ctx, err)
+		if len(accessRecords) == 0 || err != ErrTokenRequired {
+			return nil, ac.wrapErr(ctx, err)
+		}
 	}
 
 	osClient, err := ac.registryClient.ClientFromToken(bearerToken)
@@ -249,11 +251,17 @@ func (ac *AccessController) Authorized(ctx context.Context, accessRecords ...reg
 					continue
 				}
 				if err := verifyPruneAccess(ctx, osClient); err != nil {
+					if len(bearerToken) == 0 && err == ErrOpenShiftAccessDenied {
+						return nil, ac.wrapErr(ctx, ErrTokenRequired)
+					}
 					return nil, ac.wrapErr(ctx, err)
 				}
 				verifiedPrune = true
 			default:
 				if err := verifyImageStreamAccess(ctx, imageStreamNS, imageStreamName, verb, osClient); err != nil {
+					if len(bearerToken) == 0 && err == ErrOpenShiftAccessDenied {
+						return nil, ac.wrapErr(ctx, ErrTokenRequired)
+					}
 					if access.Action != "pull" {
 						return nil, ac.wrapErr(ctx, err)
 					}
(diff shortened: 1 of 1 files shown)


pkg/dockerregistry/server/auth.go
```diff
@@ -248,12 +250,18 @@ func (ac *AccessController) Authorized(ctx context.Context, accessRecords ...reg
...
+				}
 				if err := verifyPruneAccess(ctx, osClient); err != nil {
 					return nil, ac.wrapErr(ctx, err)
 				}
 				verifiedPrune = true
 			default:
 				if err := verifyImageStreamAccess(ctx, imageStreamNS, imageStreamName, verb, osClient); err != nil {
+					if len(bearerToken) == 0 && err == ErrOpenShiftAccessDenied {
+						return nil, ac.wrapErr(ctx, ErrTokenRequired)
+					}
```
bparees: why is prune handled different from default?  Just because you know prune will never be allowed without a valid token?

bparees:
/hold

still discussing with @smarterclayton whether this is needed, i'm not sure it is.

pkg/dockerregistry/server/auth.go
legionus: for pruning, we check the permissions to delete images (verifyPruneAccess) because `/admin/blobs` is for deleting the blobs itself and not its layer links.

The `default` case is for all other requests that must be made within repository (imagestream).

bparees: my question is:  why don't you invoke "verifyPruneAccess" with whatever token you have (and allow it to be rejected if the token is empty), just like you do for "verifyImageStreamAccess" ?

pkg/dockerregistry/server/auth.go
```diff
@@ -262,6 +270,9 @@ func (ac *AccessController) Authorized(ctx context.Context, accessRecords ...reg
 			}
 
 		case "signature":
+			if len(bearerToken) == 0 {
+				return nil, ac.wrapErr(ctx, ErrTokenRequired)
+			}
```
bparees: these are the cases i was talking about this morning, why do you not also allow these operations to be attempted when no token is provided?

if we're going to do this, the implementation should be such that for *all* api invocations, no token is required and the call succeeds if anonymous permissions are sufficient.

pkg/dockerregistry/server/auth.go
```diff
@@ -248,12 +250,18 @@ func (ac *AccessController) Authorized(ctx context.Context, accessRecords ...reg
 				if verifiedPrune {
 					continue
 				}
+				if len(bearerToken) == 0 {
+					return nil, ac.wrapErr(ctx, ErrTokenRequired)
+				}
```
bparees: why do you not make the verify call, in case anonymous access is sufficient?  (yes i know that is extremely unlikely, but we should be consistent across all api invocations)

legionus:
@bparees Comments addressed.

pkg/dockerregistry/server/auth.go
```diff
@@ -310,13 +333,19 @@ func (ac *AccessController) Authorized(ctx context.Context, accessRecords ...reg
 					return nil, ac.wrapErr(ctx, ErrUnsupportedAction)
 				}
 				if err := verifyCatalogAccess(ctx, osClient); err != nil {
+					if len(bearerToken) == 0 && err == ErrOpenShiftAccessDenied {
+						return nil, ac.wrapErr(ctx, ErrTokenRequired)
+					}
 					return nil, ac.wrapErr(ctx, err)
 				}
 			default:
 				return nil, ac.wrapErr(ctx, ErrUnsupportedResource)
```
bparees: why doesn't this one have a bearer token length check and return ErrTokenRequired?

legionus: If you made a request without a token or with an anonymous token and got EPERM, then you need to get the correct token. Otherwise, it is a real EPERM.

legionus: @bparees I do not return EPERM here only if there is no bearer. I redirect the user for the token. The token endpoint will [create a bearer token](https://github.com/openshift/image-registry/blob/master/pkg/dockerregistry/server/token.go#L54) even if you did not have it and next time I will return EPERM.

bparees: ErrUnsupportedResource doesn't look like a redirect to get a token?

legionus: If you talking about `default` case, then this is a handling of the situation when the user came with the wrong scope. This is not the absence of a token but an error.

legionus: I did not immediately understand that you are not asking about the code that I changed, but about the code from the context.

PR #129, opened 2018-10-10
[WIP] Do not store layer links in the repository

legionus:
There are several ways to lose links in the repository. User can delete imagestreamtag or whole imagestream from etcd. So the `oc prune image` will not help here (only hardprune can help when it learns to remove them). At the same time, links are involved in access control. To avoid possible harm, we must always use the information from the etcd to resolve links. We have a global descriptor cache so we will not constantly access to master api.

### DIFF
--- a/pkg/dockerregistry/server/app.go
+++ b/pkg/dockerregistry/server/app.go
@@ -17,6 +17,7 @@ import (
 	"github.com/openshift/image-registry/pkg/dockerregistry/server/cache"
 	"github.com/openshift/image-registry/pkg/dockerregistry/server/client"
 	registryconfig "github.com/openshift/image-registry/pkg/dockerregistry/server/configuration"
+	"github.com/openshift/image-registry/pkg/dockerregistry/server/linklessdriver"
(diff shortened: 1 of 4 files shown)


bparees:
doesn't this break pulling? (aren't the links in storage needed to properly pull blobs?)

legionus:
>  Otherwise it can give us considerable performance degradation in the basic scenario: push an image to a repository for one node (builders) and pull it from another node (workers).

The registry will make extra requests per node to fetch whole imagestream and will cache all descriptors.

dmage:
The problem is the mapping ImageStream -> Image, which gives us additional O(n) requests. And we can't cache it, because it's dynamic and can be changed with a simple `oc tag` command.

legionus:
I do not argue that with the new API everything will be easier.

dmage:
Err: we can't cache it, but we cache it... :(

legionus:
We have bugs on a lot of sprints ahead ))

legionus:
This PR is broken by design. We cannot stop creating links without additional objects in the database.

PR #130, opened 2018-10-11
Remove digest from cache then we remove blob

legionus:


pkg/dockerregistry/server/admin.go
```diff
@@ -81,9 +84,14 @@ func (bh *blobHandler) Delete(w http.ResponseWriter, req *http.Request) {
 		return
 	}
 
+	err := bh.Cache.Remove(bh.Digest)
+	if err != nil {
+		dcontext.GetLogger(bh).Errorf("blobHandler: ignore error: unable to remove %q from cache: %v", bh.Digest, err)
```
bparees: is this going to log an error in cases where the digest simply didn't exist in the cache?

bparees: @legionus please be sure to answer this question ^^

dmage: @bparees the current cache implementation returns an error only when the digest is invalid, which should never happen.

bparees: thanks

PR #131, opened 2018-10-12
[WIP] Clean cache on the all nodes

legionus:
To make it work we need to create:
```yaml
apiVersion: v1
kind: Service
metadata:
  name: image-registry-hardprune
spec:
  clusterIP: None
  selector:
    docker-registry: default
  type: ClusterIP
```

/cc @bparees

### DIFF
--- a/pkg/dockerregistry/server/admin.go
+++ b/pkg/dockerregistry/server/admin.go
@@ -2,7 +2,12 @@ package server
 
 import (
 	"fmt"
+	"net"
 	"net/http"
+	"net/url"
+	"os"
+
+	restclient "k8s.io/client-go/rest"
 
 	"github.com/docker/distribution"
 	dcontext "github.com/docker/distribution/context"
@@ -17,6 +22,10 @@ import (
 
 	"github.com/openshift/image-registry/pkg/dockerregistry/server/api"
 	"github.com/openshift/image-registry/pkg/dockerregistry/server/cache"
+)
+
+var (
+	pruneService string = "image-registry-hardprune"
 )
 
 func (app *App) registerBlobHandler(dockerApp *handlers.App) {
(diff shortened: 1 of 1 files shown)


bparees:
what is this fixing?  just preventing you from having to restart the registries after a hard prune?

legionus:
> what is this fixing? just preventing you from having to restart the registries after a hard prune?

Now we have a cache that keeps information about the digest for [10 minutes](https://github.com/openshift/image-registry/blob/master/images/dockerregistry/config.yml#L51). Any pruners (soft and hard) that deletes blobs breaks this cache in the worst possible way. We don’t have a blob, and we tell docker clients that we have it. Therefore, it is now required to restart all the registry pods after each pruning.

This PR should solve this problem. Each server forwards requests for deletion to others.

bparees:
> Now we have a cache that keeps information about the digest for 10 minutes. 

when was that cache introduced/turned on in our registry?

legionus:
> when was that cache introduced/turned on in our registry?

about [v3.3](https://github.com/openshift/origin/blob/release-1.3/images/dockerregistry/config.yml#L32). For a long time we had several caches at the same time. One cashe inside `docker/distribution` and another in the our middleware. Upstream cache almost never cleared. I combined them and now this ttl at least always works.

pkg/dockerregistry/server/admin.go
```diff
@@ -112,5 +121,81 @@ func (bh *blobHandler) Delete(w http.ResponseWriter, req *http.Request) {
...
+		}
+	}
+
+	hostAddrs, err := net.LookupHost(hostname)
+	if err != nil {
+		dcontext.GetLogger(bh).Errorf("propagateRequest: lookup hostname error: %s", err)
+		return
+	}
+
+	addrs, err := net.LookupHost(pruneService)
```
bparees: what is creating prune-service?  I assume it's intended to be a headless service so you can get all the addressed of the endpoints?

legionus: > what is creating prune-service?

Right now no one. But if we are going to merge this, then we can create it in the `cluster-image-registry-oprator`.

> I assume it's intended to be a headless service so you can get all the addressed of the endpoints?

Yes. I want to have another headless service with our selector to get dns records for all endpoints.

PR #134, opened 2018-10-29
[WIP]check PR test job invocation

bparees:


bparees:
@dmage @legionus fyi, the registry tests passed here, though as @dmage has noted, a bunch of our tests are marked [local] and thus are not run by this job.  We will need to work to re-enable those tests somewhere, either via the VM test job mechanism, or restructuring the tests so they do not require local node access.

PR #136, opened 2018-12-20
use protobuf for client/api communication

bparees:


smarterclayton:
Looks correct but we should double check the metrics on the run

pkg/cmd/dockerregistry/dockerregistry.go
```diff
@@ -179,7 +179,10 @@ func Execute(configFile io.Reader) {
 func NewServer(ctx context.Context, dockerConfig *configuration.Configuration, extraConfig *registryconfig.Configuration) (*http.Server, error) {
 	setDefaultLogParameters(dockerConfig)
 
-	registryClient := client.NewRegistryClient(clientcmd.NewConfig().BindToFile(extraConfig.KubeConfig))
+	clientConfig := clientcmd.NewConfig()
+	clientConfig.CommonConfig.AcceptContentTypes = "application/vnd.kubernetes.protobuf,application/json"
+	clientConfig.CommonConfig.ContentType = "application/vnd.kubernetes.protobuf"
+	registryClient := client.NewRegistryClient(clientConfig.BindToFile(extraConfig.KubeConfig))
```
smarterclayton: BindToFile ignore whatever the original client config is, it looks like.

smarterclayton: It doesn't ignore it, but this is not working the way I would expect.  Still seeing json to the server in the metrics capture.

bparees: @smarterclayton hm. i just refactored it to add the content setting after the bindtofile call and pushed, but it sounds like you don't think that'll matter.

bparees:
@smarterclayton do you want to pick this up again?

smarterclayton:
Yeah I do

bparees:
> Yeah I do

@smarterclayton ok, we'll have to chat about what is missing here then since you indicated this didn't accomplish the goal as i recall.

unless you just want to merge this and see what happens, which i'm ok with too.

/retest

bparees:
ref from @mfojtik for whenever this gets picked up again:
https://github.com/openshift/library-go/blob/69ec53fc354e648d1b1bd08556e702639f494688/pkg/operator/staticpod/certsyncpod/certsync_cmd.go#L107

PR #137, opened 2018-12-20
WIP - Example service monitors

smarterclayton:
Just based on debugging locally.  These would need to change a bit for 4.0 but at least cover the "obvious mistakes" part.

### DIFF
--- a/hack/service-monitor.yaml
+++ b/hack/service-monitor.yaml
@@ -1,0 +1,24 @@
+apiVersion: monitoring.coreos.com/v1
+kind: ServiceMonitor
+metadata:
+  labels:
+    k8s-app: image-registry
+  name: image-registry
+  namespace: openshift-monitoring
+spec:
+  endpoints:
+  - bearerTokenFile: /var/run/secrets/kubernetes.io/serviceaccount/token
+    interval: 30s
+    port: "5000-tcp"
+    path: /extensions/v2/metrics
+    scheme: https
+    targetPort: 5000
+    tlsConfig:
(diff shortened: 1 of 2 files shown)


bparees:
is there a reason not to make this a PR against the cluster-image-registry-operator repo instead?

hack/service-monitor.yaml
```diff
@@ -0,0 +1,24 @@
...
+spec:
+  endpoints:
+  - bearerTokenFile: /var/run/secrets/kubernetes.io/serviceaccount/token
+    interval: 30s
+    port: "5000-tcp"
+    path: /extensions/v2/metrics
+    scheme: https
+    targetPort: 5000
+    tlsConfig:
+      caFile: /var/run/secrets/kubernetes.io/serviceaccount/service-ca.crt
```
bparees: is the service-monitor pod somehow ensuring this gets mounted since the auth controller doesn't create this anymore? (or won't be past 4.0, but i think it's also currently missing)

hack/service-monitor.yaml
```diff
@@ -0,0 +1,24 @@
...
+      serverName: docker-registry.default.svc
```
bparees: image-registry.openshift-image-registry.svc

(though we'll have to sort out migration implications as well)

smarterclayton:
Just a sample, I wanted you to have it handy even if it only works for
3.11.  Will close when someone actually has time to look.

bparees:
thanks, i've linked this PR in https://jira.coreos.com/browse/DEVEXP-152

PR #138, opened 2018-12-21
Reduce cardinality of metrics, use summary instead of histograms

smarterclayton:
The repo latency metrics could have extremely high cardinality, so without
a definite use case let's remove it. Instead of Histograms, use Summaries
which simplify buckets and make visualization easier.

### DIFF
--- a/pkg/dockerregistry/server/metrics/prometheus.go
+++ b/pkg/dockerregistry/server/metrics/prometheus.go
@@ -18,13 +18,13 @@ const (
 )
 
 var (
-	requestDurationSeconds = prometheus.NewHistogramVec(
-		prometheus.HistogramOpts{
+	requestDurationSeconds = prometheus.NewSummaryVec(
+		prometheus.SummaryOpts{
 			Namespace: namespace,
 			Name:      "request_duration_seconds",
 			Help:      "Request latency in seconds for each operation.",
 		},
-		[]string{"operation", "name"},
+		[]string{"operation"},
 	)
 
 	HTTPInFlightRequests = prometheus.NewGauge(prometheus.GaugeOpts{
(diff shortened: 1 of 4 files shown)


pkg/dockerregistry/server/metrics/prometheus.go
```diff
@@ -42,43 +42,39 @@ var (
...
 		[]string{"code", "method"},
 	)
-	HTTPRequestDurationSeconds = prometheus.NewHistogramVec(
-		prometheus.HistogramOpts{
+	HTTPRequestDurationSeconds = prometheus.NewSummaryVec(
+		prometheus.SummaryOpts{
 			Namespace: namespace,
 			Subsystem: httpSubsystem,
 			Name:      "request_duration_seconds",
 			Help:      "A histogram of latencies for requests to the registry.",
```
bparees: help text is wrong now

pkg/dockerregistry/server/metrics/prometheus.go
```diff
@@ -42,43 +42,39 @@ var (
...
 		[]string{"method"},
 	)
-	HTTPRequestSizeBytes = prometheus.NewHistogramVec(
-		prometheus.HistogramOpts{
+	HTTPRequestSizeBytes = prometheus.NewSummaryVec(
+		prometheus.SummaryOpts{
 			Namespace: namespace,
 			Subsystem: httpSubsystem,
 			Name:      "request_size_bytes",
 			Help:      "A histogram of sizes of requests to the registry.",
```
bparees: help text is wrong

pkg/dockerregistry/server/metrics/prometheus.go
```diff
@@ -42,43 +42,39 @@ var (
...
 		[]string{},
 	)
-	HTTPResponseSizeBytes = prometheus.NewHistogramVec(
-		prometheus.HistogramOpts{
+	HTTPResponseSizeBytes = prometheus.NewSummaryVec(
+		prometheus.SummaryOpts{
 			Namespace: namespace,
 			Subsystem: httpSubsystem,
 			Name:      "response_size_bytes",
 			Help:      "A histogram of response sizes for requests to the registry.",
```
bparees: help text is wrong

pkg/dockerregistry/server/metrics/prometheus.go
```diff
@@ -42,43 +42,39 @@ var (
...
 		[]string{},
 	)
-	HTTPTimeToWriteHeaderSeconds = prometheus.NewHistogramVec(
-		prometheus.HistogramOpts{
+	HTTPTimeToWriteHeaderSeconds = prometheus.NewSummaryVec(
+		prometheus.SummaryOpts{
 			Namespace: namespace,
 			Subsystem: httpSubsystem,
 			Name:      "time_to_write_header_seconds",
 			Help:      "A histogram of request durations until the response headers are written.",
```
bparees: help text is wrong

pkg/dockerregistry/server/metrics/prometheus.go
smarterclayton: a summary is a histogram

bparees: hm.  prometheus docs treat them as two different things with slightly different behaviors:

https://prometheus.io/docs/practices/histograms/

but ok.

smarterclayton:
Fixed the test

dmage:
But Summaries are not aggegatable...

pkg/dockerregistry/server/metrics/prometheus.go
dmage: Not ok, math treats them differently too. A summary is not a histogram, it is a quantile.

PR #140, opened 2019-01-08
install rsync in docker-registry repo

bparees:
https://bugzilla.redhat.com/show_bug.cgi?id=1664854


Dockerfile.rhel7
```diff
@@ -4,6 +4,7 @@ COPY . .
 RUN hack/build-go.sh
 
 FROM registry.svc.ci.openshift.org/ocp/4.0:base
+RUN yum install -y rsync
```
bparees: @smarterclayton is this an allowed dockerfile change (and will it work w/ our build system?)

PR #143, opened 2019-01-22
Allow DELETE for incomplete uploads

dmage:


bparees:
I think we'll need to backport this one to 3.11 for the customer issue

PR #150, opened 2019-01-24
[3.9] Fix gometalinter in Travis

dmage:


bparees:
looks like we have no real CI coverage beyond travis for the 3.9 branch, probably because there's no 3.9 here:

https://github.com/openshift/release/tree/master/ci-operator/config/openshift/image-registry

PR #152, opened 2019-01-30
clean up our image building process

bparees:
1) remove an old/now unused dockerfile that was causing confusion

2) update our binary build process to always output to a single dir regardless of platform, this was requested by the ART team.  I don't think we care about cross-platform building any more.


bparees:
@sosiouxme fyi

bparees:
@dmage ok, i've updated this to at least work... it's pretty hacky so if you have a better suggestion i'm certainly open to it, but for now i think this gets us where we need to be... ultimately we can probably revisit all these built scripts since we don't have a need for cross-compilation any more (or at least so i assume...  none of our operators are being cross-compiled currently).

dmage:
`failed to get a token for the user alice: the server could not find the requested resource (post users.user.openshift.io)`
looks like the same problem as in https://github.com/openshift/cluster-image-registry-operator/pull/180

/lgtm
/retest

PR #156, opened 2019-02-06
Make metrics aggregatable

dmage:
This PR basically partially reverts 89585b5261d8ac4df6ec13a90a5d95100ceaa732.

We want our metrics to be aggregatable if we have more than one replica. The summary type might be convenient for one replica, but it's useless if we have more replicas.

See https://prometheus.io/docs/practices/histograms/ and http://latencytipoftheday.blogspot.com/2014/06/latencytipoftheday-you-cant-average.html

/cc @smarterclayton @bparees

### DIFF
--- a/pkg/dockerregistry/server/metrics/prometheus.go
+++ b/pkg/dockerregistry/server/metrics/prometheus.go
@@ -18,8 +18,8 @@ const (
 )
 
 var (
-	requestDurationSeconds = prometheus.NewSummaryVec(
-		prometheus.SummaryOpts{
+	requestDurationSeconds = prometheus.NewHistogramVec(
+		prometheus.HistogramOpts{
 			Namespace: namespace,
 			Name:      "request_duration_seconds",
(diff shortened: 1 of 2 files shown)


smarterclayton:
Why is summary useless?  We “max by (instance)” all the time on other multiple instance things like the kubeapiserver, which is also multi instance.  Summary is way easier to use as an admin than the alternative.

smarterclayton:
I would always use a summary over histogram for normal use.

smarterclayton:
The best argument is really how much easier it is to see latency, vs having to continually build a histogram via query.  When you need the info you need it fast

dmage:
> We “max by (instance)” all the time on other multiple instance things

This works only when your requests are evenly distributed and all nodes have the same performance. API requests on our CI cluster is distributed like (33%, 66%, 0%). If we have a node that handles a 1% of requests, but this node is slow, it will spoil all the metrics.

But as you wish.

PR #158, opened 2019-02-20
Add request tracer to avoid loops 

legionus:


bparees:
@legionus ok the easy way failed, can you go ahead and backport this to 3.11?

thanks

PR #160, opened 2019-02-20
[3.11] Don't use pull through for public registry name

dmage:


pkg/imagestream/imagestream.go
```diff
@@ -267,21 +267,37 @@ func (is *imageStream) Exists(ctx context.Context) (bool, *rerrors.Error) {
...
+			fmt.Sprintf("localRegistry: unable to parse dockerImageRepository %q", stream.Status.DockerImageRepository),
+			err,
+		)
+	} else if len(local.Registry) != 0 {
+		localNames = append(localNames, local.Registry)
+	}
+
+	public, err := imageapi.ParseDockerImageReference(stream.Status.PublicDockerImageRepository)
+	if err != nil {
+		return nil, rerrors.NewError(
```
bparees: i think you should still return localNames here.  The caller of localRegistry() isn't even checking the error (which should perhaps be fixed) and we at least want to filter the internal hostname, even if we can't parse the external one

dmage:
#161

PR #161, opened 2019-02-20
[3.11] prevent external hostname pullthrough + bump digest cache size

bparees:


pkg/dockerregistry/server/app.go
```diff
@@ -23,8 +23,8 @@ import (
 
 const (
 	// Default values
-	defaultDescriptorCacheSize         = 4096
-	defaultDigestToRepositoryCacheSize = 2048
+	defaultDescriptorCacheSize         = 6 * 4096
+	defaultDigestToRepositoryCacheSize = 6 * 2048
```
smarterclayton: I don't think you have to bump this, this is per digest anyway.

pkg/dockerregistry/server/app.go
```diff
@@ -23,8 +23,8 @@ import (
 
 const (
 	// Default values
-	defaultDescriptorCacheSize         = 4096
-	defaultDigestToRepositoryCacheSize = 2048
+	defaultDescriptorCacheSize         = 6 * 4096
```
smarterclayton: Add a comment about the rough size estimate ~16Mb of cache

pkg/imagestream/imagestream.go
```diff
@@ -267,21 +267,31 @@ func (is *imageStream) Exists(ctx context.Context) (bool, *rerrors.Error) {
...
+	var localNames []string
+
 	local, err := imageapi.ParseDockerImageReference(stream.Status.DockerImageRepository)
 	if err != nil {
-		return "", rerrors.NewError(
-			ErrImageStreamUnknownErrorCode,
-			fmt.Sprintf("localRegistry: unable to parse reference %q", stream.Status.DockerImageRepository),
-			err,
-		)
+		glog.Warningf("localRegistry: unable to parse dockerImageRepository %q", stream.Status.DockerImageRepository)
```
dmage: We don't use glog, we use `dcontext.GetLogger(ctx).Warnf`

bparees: thanks, fixed.

PR #166, opened 2019-02-26
[3.11] use new docker-distribution branch

dmage:
This commit changes the branch for Distribution and pins dependencies.

k8s.io/* are not pinned, the only change from them is removal of vendor/k8s.io/.../BUILD files.


glide.yaml
```diff
@@ -28,27 +28,26 @@ import:
 
 # openshift third
 - package: github.com/openshift/api
-  version: master
+  version: 072d1972d82ada59b68aed5e7c6ba93685edeb78
 - package: github.com/openshift/client-go
-  version: master
+  version: 5b198a036f098f34fbf0c42768382d80708cd2a9
```
bparees: why not pin these to the 3.11 branches?

dmage: yes, we can use release-3.11, updated.

PR #169, opened 2019-03-07
Use kubernetes-1.12.6

dmage:


bparees:
seems like we're safer staying on the level that origin is using (especially in case of carries), what's the motivation to move off it?

dmage:
It seems that [other components](https://github.com/openshift/cluster-openshift-apiserver-operator/blob/master/glide.yaml) use the upstream branches.

bparees:
/lgtm 

but we should get back on the origin fork when we can.

bparees:
> It seems that other components use the upstream branches.

it's not the most critical thing, but it seems like we'd benefit from using a common level where possible (ie where we don't have unique carries)

PR #174, opened 2019-04-12
Use a static libtrust key to ensure consistent pullthrough manifests

smarterclayton:
The signing key in a schema1 manifest is effectively unused or verified,
and no client leverages them today. However, the use of a randomly generated
key ensures that subsequent pulls of a schema1 manifest via the v2 endpoint
will always return different body content, which defeats the ability of a
caller to verify the content they pulled.

Instead of generating every time, hardcode a specific key so that the body
content is stable and predictable. When schema1 support is removed the key
will be inert.

### DIFF
--- a/pkg/dockerregistry/server/configuration/configuration.go
+++ b/pkg/dockerregistry/server/configuration/configuration.go
@@ -13,7 +13,7 @@ import (
 	"time"
 
 	log "github.com/sirupsen/logrus"
-	"gopkg.in/yaml.v2"
+	yaml "gopkg.in/yaml.v2"
 
 	//"github.com/docker/distribution/registry/auth"
 	"github.com/docker/distribution/configuration"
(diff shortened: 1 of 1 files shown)


smarterclayton:
Triggered by https://github.com/openshift/origin/pull/22558, this is one option to ensure we provide a consistent pull through option (can pull a digest multiple times).

pkg/dockerregistry/server/configuration/configuration.go
```diff
@@ -539,8 +539,38 @@ func migrateMiddleware(dockercfg *configuration.Configuration, cfg *Configuratio
 	return nil
 }
 
+// staticTrustKey is provided to ensure all registries serve stable schema1
+// manifests when importing / proxying from schema1 registries. Since signing
+// keys in schema1 are unused and unverified, the static key here provides only
+// a guarantee that the manifests pulled from this registry have a consistent
+// digest when pulled sequentially.
```
bparees: s/pulled sequentially/repulled subsequently/ ?

smarterclayton:
This is actually not needed, I forgot canonical digests are calculated from a different subfield on schema1 manifests.  So was user error.

PR #183, opened 2019-07-18
Bug 1729979: make the registry the pid 1 process

dmage:


dmage:
It is blocked on openshift/release#4446.

runcom:
:thinking:

PR #187, opened 2019-08-09
Bug 1739724: Update make verify

coreydaley:
Bring make verify in line with what is used in newer repos
such as cluster-image-registry-operator in preparation for
discontinuing use of travisci


Makefile
```diff
@@ -52,17 +52,13 @@ check: verify test-unit test-integration
 #
 # Example:
 #   make verify
-verify: verify-fmt verify-govet verify-imports verify-gometalinter verify-sec
+verify: verify-fmt verify-imports verify-sec
```
bparees: why are we dropping govet and gometalinter?  do we not feel they add value? are they causing problems?

coreydaley: gometalinter is deprecated and being replaced with golangci-lint (which throws a ton of linting errors when run) and will take some time to fix all of them so that it will pass.

golangci-lint includes a staticcheck that is "go vet on steroids" so i figured we would use that instead.

coreydaley:
@bparees ptal, updated per our conversation

bparees:
/lgtm

@coreydaley to followup with adding golangci-lint to replace the removed go vet

coreydaley:
> /lgtm
> 
> @coreydaley to followup with adding golangci-lint to replace the removed go vet

already added an issue for that

PR #192, opened 2019-08-21
Bug 1710502: Implementing graceful shutdown.

ricardomaraschini:
Running webserver in a separated go routine and awaiting for SIGTERM.

### DIFF
--- a/pkg/cmd/dockerregistry/dockerregistry.go
+++ b/pkg/cmd/dockerregistry/dockerregistry.go
@@ -10,7 +10,9 @@ import (
 	"io/ioutil"
 	"net/http"
 	"os"
+	"os/signal"
 	"strings"
+	"syscall"
 	"time"
 
 	logrus_logstash "github.com/bshuster-repo/logrus-logstash-hook"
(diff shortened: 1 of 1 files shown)


pkg/cmd/dockerregistry/dockerregistry.go
```diff
@@ -164,14 +166,25 @@ func Execute(configFile io.Reader) {
...
+			dcontext.GetLogger(ctx).Infof("listening on %s", srv.Addr)
+			errc <- srv.ListenAndServe()
+		} else {
+			dcontext.GetLogger(ctx).Infof("listening on %s, tls", srv.Addr)
+			errc <- srv.ListenAndServeTLS(dockerConfig.HTTP.TLS.Certificate, dockerConfig.HTTP.TLS.Key)
+		}
+	}()
+
+	<-sigs
+	if err := srv.Close(); err != nil {
```
dmage: You can use Shutdown for a graceful shutdown.

pkg/cmd/dockerregistry/dockerregistry.go
```diff
@@ -164,16 +166,36 @@ func Execute(configFile io.Reader) {
...
+	select {
+	case <-sigs:
+	case err := <-errc:
+		log.Fatal(err)
 	}
-	if err != nil {
+
+	// signal has been received, attempt to stop the server within the
+	// default termination grace period of 30 seconds.
+	ctx, cancel := context.WithTimeout(ctx, 29*time.Second)
```
dmage: > period of 30 seconds

> `29*time.Second`

Why are the numbers different?

ricardomaraschini: The grace period is 30 seconds, i was planning to give it 1 entire second between receiving the signal and calling Shutdown(), maybe i could update the comment, what do you think?

dmage: Yes, it worth documenting this difference so that the next reader won't be surprised.

ricardomaraschini: done.

ricardomaraschini:
> /hold
> 
> We need an e2e for this, unfortunately the registry operator is the best place to do it. Test should do the following:
> 
> 1. Make sure a registry is available
> 2. Update the operator to scale the registry down to 0
> 3. Wait 30 seconds
> 4. Verify there are no registry pods.
> 
> I'll cancel the hold once the e2e PR is up.

Implemented(with a slightly different approach) on https://github.com/openshift/cluster-image-registry-operator/pull/372

PR #194, opened 2019-09-16
Bug 1749256: Checking if image name is valid, prune otherwise.

ricardomaraschini:
In some cases we have some images on the blob storage that contain names
considered to be invalid by the ImageStream API. Names containing / or
even % are not considered valid for Images.

As some customers are not able to prune their images due to this error
this patch adds a check and sets the image to be pruned if it contains
an invalid name(they should not be there anyways).

### DIFF
--- a/pkg/dockerregistry/server/prune/prune.go
+++ b/pkg/dockerregistry/server/prune/prune.go
@@ -14,6 +14,7 @@ import (
 
 	kerrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/rest"
 
 	dockerapiv10 "github.com/openshift/api/image/docker10"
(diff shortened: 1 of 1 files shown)


adambkaplan:
/hold

Until 4.3 is open for merges.

ricardomaraschini:
I am still going through the bugzilla process, keeping this on hold.

dmage:
/retitle Bug 1749256: Checking if image name is valid, prune otherwise.

ricardomaraschini:
@adambkaplan @dmage I am going to cherry pick this up to 3.11 as its life cycle spans up to Apr 2022. IIUC version 3.10 support period end on next 1st and as this is not a blocking bug(we can still prune images manually) I believe we are in a good shape.

ricardomaraschini:
We need to wait for #195 or alternative fix to go in.

ricardomaraschini:
As #195 was closed we are not depending on #196

PR #195, opened 2019-09-26
Disabled pprof and marking exec() as safe.

ricardomaraschini:
gosec is complaining about two issues. One is that we have pprof enabled
on our process and the other is about using variables when shelling out
commands. As far as I have checked there is no potential harm in using
exec the way we are using as command line arguments are properly escaped
before shelling out.

As for pprof, this patch disables it.

This is to fix problems during CI as this:

https://prow.svc.ci.openshift.org/view/gcs/origin-ci-test/pr-logs/pull/openshift_image-registry/194/pull-ci-openshift-image-registry-master-verify/11

### DIFF
--- a/cmd/dockerregistry/main.go
+++ b/cmd/dockerregistry/main.go
@@ -5,7 +5,6 @@ import (
 	"fmt"
 	"math/rand"
 	"net/http"
-	_ "net/http/pprof"
 	"os"
 	"runtime"
 	"time"
--- a/tools/changelog/changelog.go
+++ b/tools/changelog/changelog.go
@@ -85,6 +85,7 @@ func main() {
 	from := os.Args[1]
 	to := os.Args[2]
 
+	// #nosec
 	out, err := exec.Command("git", "log", "--topo-order", "--pretty=tformat:%h %p|%s", "--reverse", fmt.Sprintf("%s..%s", from, to)).CombinedOutput()
 	if err != nil {
 		log.Fatal(err)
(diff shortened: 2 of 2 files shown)


cmd/dockerregistry/main.go
```diff
@@ -5,7 +5,6 @@ import (
 	"fmt"
 	"math/rand"
 	"net/http"
-	_ "net/http/pprof"
```
dmage: This will break the `startProfiler` function.

tools/changelog/changelog.go
```diff
@@ -85,6 +85,7 @@ func main() {
 	from := os.Args[1]
 	to := os.Args[2]
 
+	// #nosec
```
dmage: There is no reason to analyze tools/ for security problems. It's not a part of production code.

cmd/dockerregistry/main.go
bparees: though i don't know that we've had a use for it so far, i'd be reluctant to ship code that we can't profile at a customer site if need be.  is there a way to make this configurable?

ricardomaraschini: It seems to me that I haven't done the due diligence on this one. Let me take a look.

tools/changelog/changelog.go
ricardomaraschini: Well, it is part of our codebase nonetheless.

dmage: I'm almost positive we can remove this tool from our codebase.

ricardomaraschini: I got thrilled now. How can we be 100% positive?

dmage: We can remove it and wait until someone notices it :)

ricardomaraschini:
Team decided to take `gosec` out of the way for now, a PR is following.

PR #203, opened 2019-10-23
Use Kubernetes 1.16.2

dmage:


dmage:
@openshift/openshift-team-developer-experience ptal

PR #207, opened 2019-11-04
Bug 1768507: keep digest when image is not referenced by tag

dmage:


dmage:
/assign @ricardomaraschini 
/assign @coreydaley 

https://github.com/openshift/image-registry/blob/656065ee66b798f06b8f8f3552ba8de1b4750a94/pkg/dockerregistry/server/pullthroughmanifestservice.go#L68

This line was changing the reference in the error message.

PR #209, opened 2019-12-05
Kubernetes 1.17.0

dmage:


Makefile
```diff
@@ -138,14 +138,11 @@ build-images: build-rpms
...
-	go get -d -u k8s.io/api@kubernetes-1.16.2 \
-		k8s.io/apimachinery@kubernetes-1.16.2 \
-		k8s.io/apiserver@kubernetes-1.16.2 \
-		k8s.io/client-go@kubernetes-1.16.2 \
+	go get -d -u \
 		github.com/openshift/api@release-4.3 \
 		github.com/openshift/client-go@release-4.3 \
 		github.com/openshift/library-go@release-4.3
-	go get -u=patch ./cmd/... ./pkg/... ./test/... 'github.com/docker/go-metrics@v0.0.0-20180209012529-399ea8c73916'
+	go get -u=patch ./cmd/... ./pkg/... ./test/...
```
adambkaplan: @dmage do we want to bump the openshift branches to `release-4.4`?

go.mod
```diff
@@ -36,29 +33,26 @@ require (
...
-	gopkg.in/inf.v0 v0.9.1 // indirect
-	gopkg.in/yaml.v2 v2.2.4
-	k8s.io/api v0.0.0-20191016110408-35e52d86657a
-	k8s.io/apimachinery v0.0.0-20191004115801-a2eda9f80ab8
+	gopkg.in/yaml.v2 v2.2.7
+	k8s.io/api v0.17.0
+	k8s.io/apimachinery v0.17.0
 	k8s.io/apiserver v0.0.0-20191016112112-5190913f932d // indirect
-	k8s.io/client-go v0.0.0-20191016111102-bec269661e48
+	k8s.io/client-go v0.17.0
```
adambkaplan: NO MORE SHAS!!! 🎉 🎉 🎉

test/integration/metricspullthrough/metricspullthrough_test.go
```diff
@@ -138,7 +138,7 @@ func TestPullthroughBlob(t *testing.T) {
 			values: []string{`type="Miss"`},
 		},
 		{
-			name:   "imageregistry_pullthrough_repository_duration_seconds",
+			name:   "imageregistry_pullthrough_repository_duration_seconds_count",
```
adambkaplan: @dmage did we rename these metrics, or is this an artifact of the prometheus bump?

dmage: an artifact of the prometheus bump, now we need to define Objectives for summaries. I've copied old defaults to avoid behaviour changes.

Makefile
dmage: done

pkg/dockerregistry/server/metrics/prometheus.go
```diff
@@ -20,9 +20,10 @@ const (
 var (
 	requestDurationSeconds = prometheus.NewSummaryVec(
 		prometheus.SummaryOpts{
-			Namespace: namespace,
-			Name:      "request_duration_seconds",
-			Help:      "Request latency in seconds for each operation.",
+			Namespace:  namespace,
+			Name:       "request_duration_seconds",
+			Help:       "Request latency in seconds for each operation.",
+			Objectives: map[float64]float64{0.5: 0.05, 0.9: 0.01, 0.99: 0.001},
```
adambkaplan: @dmage For my own education - are there docs on what this map does?

adambkaplan: Found them:

```
        // Objectives defines the quantile rank estimates with their respective
	// absolute error. If Objectives[q] = e, then the value reported for q
	// will be the φ-quantile value for some φ between q-e and q+e.  The
	// default value is an empty map, resulting in a summary without
	// quantiles.
```

So with this configuration we have the 50% +/- 5%, 90% +/- 1%, and 99% +/- 0.1%

adambkaplan: @dmage Was the 50/90/99 reporting previous behavior, or is this new?

dmage: @adambkaplan  it's the old default

gabemontero: fancy math / statistics :-)

PR #220, opened 2020-02-12
Bug 1745743: Implement retries when checking for blob presence.

ricardomaraschini:
This PR refers to https://bugzilla.redhat.com/show_bug.cgi?id=1745743.
We do a best effort here and retry when checking if a blob is present or
not. We need to do this as some of image registry storage backends have
eventual consistency and in some cases we may ended up checking for a
blob presence before it shows up.

This PR acts only on "manifest upload" as it tends to happen right
after the client uploaded blob layers. In the future we gonna need to
get rid of these retries, maybe by leveraging docker distribution redis
cache layer.

### DIFF
--- a/pkg/dockerregistry/server/manifesthandler/manifestschema1handler.go
+++ b/pkg/dockerregistry/server/manifesthandler/manifestschema1handler.go
@@ -4,12 +4,14 @@ import (
 	"context"
 	"fmt"
 	"path"
+	"time"
 
 	"github.com/docker/distribution"
 	"github.com/docker/distribution/manifest/schema1"
 	"github.com/docker/distribution/reference"
 	"github.com/docker/libtrust"
 	"github.com/opencontainers/go-digest"
+	"k8s.io/apimachinery/pkg/util/wait"
 
 	imageapiv1 "github.com/openshift/api/image/v1"
 	imageapi "github.com/openshift/image-registry/pkg/origin-common/image/apis/image"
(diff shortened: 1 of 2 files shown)


dmage:
/approve

This should mitigate the problem, but specially crafted requests will take 3+ seconds (and that's basically the point of this fix, to wait up to 3 seconds). I'd like to have a second opinion about this solution.

pkg/dockerregistry/server/manifesthandler/manifestschema2handler.go
```diff
@@ -80,8 +82,33 @@ func (h *manifestSchema2Handler) verifyLayer(ctx context.Context, fsLayer distri
...
 	}
 
-	desc, err := h.blobStore.Stat(ctx, fsLayer.Digest)
-	if err != nil {
+	// XXX https://bugzilla.redhat.com/show_bug.cgi?id=1745743
+	// Due to the "eventual consistent" nature of some of image registry
+	// storage driver backends we have to do here a few retries. We gonna
+	// need to address this properly in the future. Here we may take up to
+	// three seconds to report a blob as unknown.
+	var desc distribution.Descriptor
```
adambkaplan: Nit - clean up the comment to be more concise. Ex:

```
// https://bugzilla.redhat.com/show_bug.cgi?id=1745743
// AWS S3 (and potentially other object stores) only have eventual consistency guarantees.
// Stat can fail here if an image layer was recently pushed.
// In the event Stat returns `ErrBlobUnknown`, retry up to 3 seconds.
```

pkg/dockerregistry/server/manifesthandler/manifestschema1handler.go
```diff
@@ -42,8 +44,32 @@ func (h *manifestSchema1Handler) statBlob(ctx context.Context, dgst digest.Diges
...
 	}
 
-	desc, err := h.blobStore.Stat(ctx, dgst)
-	if err != nil {
+	// XXX https://bugzilla.redhat.com/show_bug.cgi?id=1745743
+	// Due to the "eventual consistent" nature of some of image registry
+	// storage driver backends we have to do here a few retries. We gonna
+	// need to address this properly in the future. Here we may take up to
+	// three seconds to report a blob as unknown.
+	if err := wait.ExponentialBackoff(
```
adambkaplan: Ditto.

pkg/dockerregistry/server/manifesthandler/manifestschema2handler.go
ricardomaraschini: done.

pkg/dockerregistry/server/manifesthandler/manifestschema1handler.go
ricardomaraschini: done.

dmage:
@adambkaplan this should never happen during the normal workflow (unless there is a consistency problem that we are trying to fix), clients ensure that all blobs are uploaded. But if you send a put manifest request with random digests, it'll take 3 seconds.

PR #221, opened 2020-02-19
Dockerfile: Clear yum cache after install

smarterclayton:
Not clearing the cache increased the image size by about 50M


smarterclayton:
4.4 bug is 1804493, 4.3 is 1804494

PR #225, opened 2020-02-27
DEVEXP-515: Using node credentials.

ricardomaraschini:
Leveraging cluster installation pull secrets during image pull through.

This PR is divided into three parts:

1. We removed a package named `registryclient` from here as we already have a similar code in `openshift/library-go/pkg/image/registryclient`(differences between the two codebases were migrated from here to there and versioned packages were updated).

2. Code that wraps a `kubernetes.DockerKeyring` into a docker/distribution `auth.CredentialStore` interface was isolated into its own package. Part of this code was embedded into the removed `registryclient` package while other parts(specifically the node credentials access) were created.

3. The last piece is to glue everything together. We migrate the imports to use `registryclient` package vendored from `library-go` and also started to take node credentials into account when finding basic auth info for registries.


pkg/credentialstores/test/image-secrets.json
```diff
@@ -0,0 +1,68 @@
+{
```
coreydaley: Don't quote me on this, but I think that we prefer yaml over json for things like this.

pkg/credentialstores/secret.go
```diff
@@ -0,0 +1,91 @@
+package credentialstores
+
+import (
```
coreydaley: Imports are typically in this order:
 - standard
 - other
 - kubernetes
 - openshift
 - module

pkg/credentialstores/secret_test.go
```diff
@@ -0,0 +1,28 @@
...
+)
+
+func TestCredentialsForSecrets(t *testing.T) {
+	data, err := ioutil.ReadFile("test/image-secrets.json")
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	var secrets corev1.SecretList
+	if err := json.Unmarshal(data, &secrets); err != nil {
```
dmage: https://github.com/openshift/library-go/blob/6d930b1c27cce8ffa64ad7982b8bc5314ee569b1/pkg/operator/resource/resourceread/core.go#L28

pkg/credentialstores/common.go
```diff
@@ -0,0 +1,129 @@
...
+	"net/url"
+	"strings"
+
+	"github.com/openshift/image-registry/pkg/kubernetes-common/credentialprovider"
+
+	"github.com/golang/glog"
+)
+
+// MultiError is a wrap for multiple errors.
+type MultiError struct {
```
coreydaley: We have used https://github.com/kubernetes/apimachinery/blob/master/pkg/util/errors/errors.go other places

pkg/credentialstores/common.go
```diff
@@ -0,0 +1,129 @@
...
+			target, regURL, "index.docker.io/v1",
+		)
+		return basicCredentialsFromKeyring(
+			keyring, &url.URL{Host: "index.docker.io", Path: "/v1"},
+		)
+	}
+
+	// docker 1.9 saves 'docker.io' in config in f23, see
+	// https://bugzilla.redhat.com/show_bug.cgi?id=1309739
+	if regURL == "index.docker.io" {
```
dmage: This logic is inside keyring.Lookup

pkg/credentialstores/common.go
```diff
@@ -0,0 +1,129 @@
...
+		glog.V(5).Infof(
+			"Found secret to match %s (%s): %s",
+			target, regURL, configs[0].ServerAddress,
+		)
+		return configs[0].Username, configs[0].Password
+	}
+
+	// do a special case check for docker.io to match historical lookups
+	// when we respond to a challenge
+	if regURL == "auth.docker.io/token" {
```
dmage: We don't support it

pkg/credentialstores/common.go
```diff
@@ -0,0 +1,129 @@
...
+
+// Error exists to comply with golang's error interface.
+func (m *MultiError) Error() string {
+	return strings.Join(m.errors, " ")
+}
+
+// basicCredentialsFromKeyring extract basicAuth information from provided
+// keyring. If keyring does not contain information for the provided URL, empty
+// strings are returned instead.
+func basicCredentialsFromKeyring(keyring credentialprovider.DockerKeyring, target *url.URL) (string, string) {
```
dmage: Why does it accept URL?

pkg/credentialstores/test/image-secrets.json
ricardomaraschini: Addressed.

pkg/credentialstores/secret.go
ricardomaraschini: Addressed.

pkg/credentialstores/secret_test.go
ricardomaraschini: I could not find a function to read a SecretList. Other than that, this code has been replaced by yaml decoding.

pkg/credentialstores/common.go
ricardomaraschini: nice catch. addressed.

pkg/credentialstores/common.go
ricardomaraschini: Because docker/distribution uses it so makes sense to keep it concise:

https://github.com/docker/distribution/blob/84f47e7bb3bb3521d7d20615d42078df306fb01d/registry/client/auth/session.go#L44-L55

pkg/credentialstores/common.go
ricardomaraschini: I don't think we can change this as this is not something like a StaticCredential.

pkg/credentialstores/common.go
dmage: DockerKeyring.Lookup accepts an image reference, not an URL. When a user provides you an image reference, you convert it into an URL, and then convert it back to an image reference-like string - you are doing something wrong.

CredentialStore is not supposed to be used that way, and we should never use it that way. You can find examples how to use it in the docker/distribution repostory.

pkg/credentialstores/common.go
dmage: Why is this not something like a StaticCredential?

ricardomaraschini: Registry client in library-go is designed around a Context entity. This entity is capable of, by means of a [CredentialStore](https://github.com/openshift/library-go/blob/master/pkg/image/registryclient/client.go#L57), fetch "repositories". 

Image registry uses library-go and creates a single CredentialStore containing all credentials available, that is IMO much better than creating individual Contexts with different "StaticCredentials" every time we need a Context to access a repository.

pkg/credentialstores/common.go
ricardomaraschini: `docker/distribution` uses CredentialStores and k8s has something called keyring, the first uses url.URL and the latter string.

k8s makes available to us functions to parse secrets into keyrings and to use them we need these keyrings to be available on our project, hence we need a conversion layer that wraps a keyring with multiple credentials into a CredentialStore as needed by library-go registryclient.

Could you tell me what would you do differently? I would most certainly adapt this to your input as this codes predates this PR and myself.

pkg/credentialstores/common.go
ricardomaraschini: This has relation with this "CredentialStore containing multiple credentials" topic we have been discussing in other comments.

ricardomaraschini:
I will be working in a different PR(https://github.com/openshift/image-registry/pull/226) for this one.

PR #226, opened 2020-03-10
DEVEXP-515: Leveraging installation pull secrets.

ricardomaraschini:
Leveraging cluster installation pull secrets during image pull through.


coreydaley:
@ricardomaraschini Is there any testing that could be added to verify this new code or is what we already had sufficient?

gabemontero:
> @ricardomaraschini Is there any testing that could be added to verify this new code or is what we already had sufficient?

FWIW @coreydaley @ricardomaraschini I have a commit for the samples operator that deletes its copying of the install pull secret to the openshift namespace, so as to leverage what this change acoomplishes

I could push that to a branch for @ricardomaraschini to leverage if you all would think it would help here.

go.mod
```diff
@@ -32,22 +32,24 @@ require (
...
 	github.com/sirupsen/logrus v1.4.2
+	golang.org/x/xerrors v0.0.0-20191204190536-9bdfabe68543 // indirect
 	google.golang.org/cloud v0.0.0-20151119220103-975617b05ea8 // indirect
 	gopkg.in/yaml.v2 v2.2.7
-	k8s.io/api v0.17.0
-	k8s.io/apimachinery v0.17.0
-	k8s.io/client-go v0.17.0
+	k8s.io/api v0.17.2
+	k8s.io/apimachinery v0.17.2
+	k8s.io/client-go v0.17.1
```
adambkaplan: @ricardomaraschini why the skew between k8s.io/api and k8s.io/client-go?

ricardomaraschini: well, that was not on purpose but I chosen to left it anyways :-) do you want me to revert?

ricardomaraschini:
> @ricardomaraschini Is there any testing that could be added to verify this new code or is what we already had sufficient?

I have added an unit test for it, PTAL.

go.mod
ricardomaraschini: It has to do with the library-go update:

https://github.com/openshift/library-go/blob/master/go.mod

pkg/dockerregistry/server/util_test.go
```diff
@@ -0,0 +1,178 @@
...
+	return nil
+}
+
+func Test_getImportContext(t *testing.T) {
+	originalCredDir := installCredentialsDir
+	defer func() {
+		installCredentialsDir = originalCredDir
+	}()
+
+	tmpCredDir, err := ioutil.TempDir("", "credentials")
```
dmage: Please clean up after yourself.

pkg/dockerregistry/server/util_test.go
```diff
@@ -0,0 +1,178 @@
...
+	"context"
+	"io/ioutil"
+	"net/http"
+	"os"
+	"testing"
+
+	"github.com/openshift/image-registry/pkg/dockerregistry/server/metrics"
+	imageapi "github.com/openshift/image-registry/pkg/origin-common/image/apis/image"
+	"github.com/openshift/library-go/pkg/image/registryclient"
+	corev1 "k8s.io/api/core/v1"
```
dmage: sorting

pkg/dockerregistry/server/util.go
```diff
@@ -40,19 +50,48 @@ func getNamespaceName(resourceName string) (string, string, error) {
...
 		return nil, err
 	}
-	secrets, err := secretsGetter()
+
+	if ref == nil {
+		return nil, fmt.Errorf("invalid nil image reference")
+	}
+
+	if m == nil {
+		return nil, fmt.Errorf("invalid nil metrics pullthrough")
```
dmage: I don't think it's a good practice to check pointer parameters for nil in every function, sometimes it's better to get a nil pointer dereference panic.

ricardomaraschini: I disagree, if a value may be nil functions should be ready to receive it as nil as well.

pkg/dockerregistry/server/util_test.go
ricardomaraschini: solved.

pkg/dockerregistry/server/util_test.go
ricardomaraschini: solved.

pkg/dockerregistry/server/util.go
ricardomaraschini: Nevermind, nil checks removed as requested.

PR #231, opened 2020-04-01
DEVEXP-555: Rebasing k8s.io

ricardomaraschini:


ricardomaraschini:
/retest

Error says:
```
(usually a result of a race or resource pressure. re-running the job should help)
```

ricardomaraschini:
Tests are failing, waiting for https://github.com/openshift/release/pull/8109 to get merged.

coreydaley:
Waiting on https://github.com/openshift/release/pull/8224

coreydaley:
I don't think we have a standard for line length in this repository, or openshift as a whole, but I feel like there are a lot of unnecessary line breaks in this pull request.

Maybe only break them up if there are more than three or they are really long?

ricardomaraschini:
> I don't think we have a standard for line length in this repository, or openshift as a whole, but I feel like there are a lot of unnecessary line breaks in this pull request.
> 
> Maybe only break them up if there are more than three or they are really long?

All new break lines removed. PTAL.

pkg/dockerregistry/server/auth.go
```diff
@@ -375,7 +375,7 @@ func getOpenShiftAPIToken(req *http.Request) (string, error) {
 }
 
 func verifyOpenShiftUser(ctx context.Context, c client.UsersInterfacer) (string, string, error) {
-	userInfo, err := c.Users().Get("~", metav1.GetOptions{})
+	userInfo, err := c.Users().Get(context.TODO(), "~", metav1.GetOptions{})
```
dmage, line 378: Use `ctx` instead of `context.TODO()`.

pkg/dockerregistry/server/signaturedispatcher.go
```diff
@@ -120,7 +120,9 @@ func (s *signatureHandler) Put(w http.ResponseWriter, r *http.Request) {
 	newSig := &imageapiv1.ImageSignature{Content: sig.Content, Type: sig.Type}
 	newSig.Name = sig.Name
 
-	_, err = client.ImageSignatures().Create(newSig)
+	_, err = client.ImageSignatures().Create(
```
dmage, line 123:
```suggestion
	_, err = client.ImageSignatures().Create(s.ctx, newSig, metav1.CreateOptions{})
```

pkg/dockerregistry/server/auth.go
ricardomaraschini: Good catch!

ricardomaraschini: I think that this problem happens in a lot of different places, I will fix them and submit a new commit. Please hold on.

pkg/dockerregistry/server/signaturedispatcher.go
ricardomaraschini: done.

adambkaplan:
> I don't think we have a standard for line length in this repository, or openshift as a whole, but I feel like there are a lot of unnecessary line breaks in this pull request.

Sadly @coreydaley I don't think `gofmt` enforces a line width, and there be dragons if we adopt something that strays too far from the golang ecosystem. I think the rule of thumb is start adding line breaks at 100 characters to make it readable on GitHub.

dmage:
@adambkaplan @coreydaley @ricardomaraschini we follow Go Code Review Comments, it has a note about line length: https://github.com/golang/go/wiki/CodeReviewComments#line-length

Long lines is a code small and wrapping just deodorizes them. So I'd agree with @coreydaley and prefer not to wrap lines in the middle of function calls, especially lines that are less than 100-120 characters (thanks @ricardomaraschini for fixing it).

pkg/dockerregistry/server/auth.go
ricardomaraschini: I sent a new commit that address context's usage, ptal.

PR #234, opened 2020-04-17
Bug 1804160: Checking for namespace existence

ricardomaraschini:
By design `docker/distribution` hides any error during auth under a BadRequest[1] so this is clearly not an optimal solution. We can't distinguish between a "non-existent" namespace and a kubernetes api error, in other words: we can't distinguish between a 403 and an 500, everything is a 400 FTW.

Let me know if I should try to patch upstream or I should be doing this validation somewhere else.

[1] https://github.com/docker/distribution/blob/master/registry/handlers/app.go#L844-L864

### DIFF
--- a/pkg/dockerregistry/server/auth.go
+++ b/pkg/dockerregistry/server/auth.go
@@ -8,6 +8,7 @@ import (
 	"net/url"
 	"strings"
 
+	"github.com/docker/distribution"
 	dcontext "github.com/docker/distribution/context"
 	registryauth "github.com/docker/distribution/registry/auth"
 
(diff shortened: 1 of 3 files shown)


ricardomaraschini:
/assign @dmage 

This is how it looks like on image-registry side.

dmage:
@ricardomaraschini this is going to fix a problem that happens only with erroneous requests. An additional request to the api server for every incoming http request because of this doesn't sound good.

ricardomaraschini:
There is an alternative version of this on https://github.com/openshift/image-registry/pull/235, closing this one.

PR #235, opened 2020-04-20
Bug 1804160: Permission denied if namespace does not exist

ricardomaraschini:


ricardomaraschini:
@dmage This would be a different approach, detecting the error when uploading the manifest. Let me know what you think.

dmage:
@ricardomaraschini looks good, we need a test for that

ricardomaraschini:
> @ricardomaraschini looks good, we need a test for that

on it.

ricardomaraschini:
@dmage ptal

dmage:
We may eventually refactor our integration tests into e2e tests
/lgtm

PR #236, opened 2020-04-21
Bug 1827672: Only warn if PublicDockerImageRepository is set

smarterclayton:
It is not an error for PublicDockerImageRepository to be set empty
to an image stream (it means the registry is not exposed by default).


adambkaplan:
/retitle Bug 1827672: Only warn if PublicDockerImageRepository is set

PR #249, opened 2020-09-29
Bug 1883502: Migrating from k8s.io/klog v1 to k8s.io/klog v2

ricardomaraschini:
Kubernetes 1.19 is leveraging k8s.io/klog v2. This PR updates references we still have to v1 to v2.


Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM registry.svc.ci.openshift.org/openshift/release:golang-1.13 AS builder
+FROM registry.svc.ci.openshift.org/openshift/release:golang-1.15 AS builder
```
dmage, line 1: is it something we want to backport to 4.6?

PR #250, opened 2020-10-07
Bug 1883853: Replace bitbucket.org/ww/goautoneg with its mirror

dmage:
bitbucket.org/ww/goautoneg is no longer available, so it is replaced by github.com/munnerz/goautoneg. This mirror is used by Kubernetes.


ricardomaraschini (approved):
/lgtm

PR #254, opened 2020-10-27
Workaround /etc/pki/ca-trust bug on OpenShift 4.6

dmage:


Dockerfile
```diff
@@ -7,7 +7,7 @@ FROM registry.svc.ci.openshift.org/openshift/origin-v4.0:base
 RUN yum install -y rsync && yum clean all && rm -rf /var/cache/yum
 COPY --from=builder /go/src/github.com/openshift/image-registry/_output/local/bin/dockerregistry /usr/bin/
 COPY images/dockerregistry/config.yml /
-RUN chmod a+w -R /etc/pki/ca-trust/extracted
+ADD images/dockerregistry/writable-pem.tar.gz /etc/pki/ca-trust/extracted
```
akhil-rane, line 10: @dmage I may be wrong about this but since we are now mounting to /etc/pki/ca-trust during runtime this won't work right?

dmage: @akhil-rane ADD works on a different level. Even if something is mounted there, ADD will still create a new layer with files, so these files will be added to the image ignoring the mounted file systems. These files won't be visible during build in this case.

ricardomaraschini (approved):
/lgtm

PR #259, opened 2020-12-08
Bug 1905502: Polling route until registry is available

ricardomaraschini:
This PR adds a check on the external route for the ephemeral registry right after its deployment, this way we can assure the registry is accessible from outside the cluster prior to tests execution.

Networking team uses the following script to achieve the same on their tests:

https://github.com/openshift/origin/blob/master/test/extended/router/scoped.go#L241-L271

### DIFF
--- a/pkg/testframework/registry.go
+++ b/pkg/testframework/registry.go
@@ -3,9 +3,12 @@ package testframework
 import (
 	"bytes"
 	"context"
+	"crypto/tls"
 	"encoding/json"
 	"fmt"
+	"io/ioutil"
 	"net"
+	"net/http"
 	"sync/atomic"
 	"testing"
 	"time"
@@ -467,10 +470,17 @@ func CreateEphemeralRegistry(t *testing.T, restConfig *rest.Config, namespace st
 			return false, nil
 		}
 		for _, ingress := range route.Status.Ingress {
-			if len(ingress.Host) > 0 {
-				host = ingress.Host
-				return true, nil
+			if len(ingress.Host) == 0 {
+				continue
 			}
+
+			if err := checkRoute(ingress.Host); err != nil {
+				lastErr = err
+				return false, nil
+			}
+
+			host = ingress.Host
+			return true, nil
 		}
 		lastErr = fmt.Errorf("route %s does not have ingress hosts", name)
 		return false, nil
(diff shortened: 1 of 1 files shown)


pkg/testframework/registry.go
```diff
@@ -482,3 +492,39 @@ func CreateEphemeralRegistry(t *testing.T, restConfig *rest.Config, namespace st
...
+
+	url := fmt.Sprintf("https://%s/v2/", host)
+	client := &http.Client{Transport: tr}
+	res, err := client.Get(url)
+	if err != nil {
+		return err
+	}
+	defer res.Body.Close()
+
+	if res.StatusCode == http.StatusServiceUnavailable {
```
adambkaplan, line 522: Do we care if the status code is something else, like 404?

In reviewing what the networking team does, their script only exits 0 if the OK (200) response is returned.

ricardomaraschini: Yes, you are right. Adjusted accordingly.

ricardomaraschini: Just clarifying this, for some tests the registry is deployed with basic auth in place. In such scenario a 403 return is also OK. I have adjusted the deployment to consider 200 or 403 as "the registry is good to go".

pkg/testframework/registry.go
```diff
@@ -482,3 +492,39 @@ func CreateEphemeralRegistry(t *testing.T, restConfig *rest.Config, namespace st
...
+
+	url := fmt.Sprintf("https://%s/v2/", host)
+	client := &http.Client{Transport: tr}
+	res, err := client.Get(url)
+	if err != nil {
+		return err
+	}
+	defer res.Body.Close()
+
+	if res.StatusCode != http.StatusOK && res.StatusCode != http.StatusUnauthorized {
```
adambkaplan, line 522: nit - add a comment that `http.StatusUnauthorized` indicates that the registry is operating normally.

adambkaplan (changes requested):

pkg/testframework/registry.go
ricardomaraschini: Good call. Done.

adambkaplan (approved):
/lgtm

PR #268, opened 2021-03-16
Better log message when action is not allowed

dmage:
SubjectAccessReview can be in one of three states: allowed, denied, and no opinion. Also it has the additional field EvaluationError that can indicate that some error occurred. The registry should log this information.


ricardomaraschini (approved):
/lgtm

PR #271, opened 2021-04-22
Bug 1843253: Allowing indirect pullthrough

ricardomaraschini:
Allowing pullthrough of one ImageStreamTag pointing to another
ImageStreamTag.

### DIFF
--- a/pkg/dockerregistry/server/pullthroughmanifestservice.go
+++ b/pkg/dockerregistry/server/pullthroughmanifestservice.go
@@ -8,9 +8,12 @@ import (
 	dcontext "github.com/docker/distribution/context"
 	"github.com/opencontainers/go-digest"
 
+	imageapiv1 "github.com/openshift/api/image/v1"
+
 	"github.com/openshift/image-registry/pkg/dockerregistry/server/cache"
 	"github.com/openshift/image-registry/pkg/dockerregistry/server/metrics"
 	"github.com/openshift/image-registry/pkg/errors"
+	rerrors "github.com/openshift/image-registry/pkg/errors"
 	"github.com/openshift/image-registry/pkg/imagestream"
 	imageapi "github.com/openshift/image-registry/pkg/origin-common/image/apis/image"
 )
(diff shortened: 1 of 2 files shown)


pkg/dockerregistry/server/pullthroughmanifestservice.go
```diff
@@ -44,35 +47,56 @@ func (m *pullthroughManifestService) Get(ctx context.Context, dgst digest.Digest
...
+
+	var ref imageapi.DockerImageReference
+	var image *imageapiv1.Image
+	seen := map[string]bool{}
+	for len(streams) > 0 {
+		is := streams[0]
+		streams = streams[1:]
+
+		var rErr rerrors.Error
+		if image, rErr = is.GetImageOfImageStream(ctx, dgst); rErr != nil {
```
smarterclayton, line 63: how is this checking access control?  from references are not protected by acl, so without a check this would let you access other peoples images?

ricardomaraschini: My reasoning, please correct me where I am wrong: I want to create an ImageStreamTag X pointing to another ImageStreamTag Y using Y's registry URL for that purpose (`internal-registry/ns/Y:latest`). This will fail if I don't have enough permissions to reach the registry and read Y's manifest, leaving my ImageStreamTag X in a faulty state (permission denied, i.e. no sha). If I do have enough permissions to reach the registry and read Y's manifest then X will be configured for pull-through from Y and from this point on it is assured that from X's namespace I can access Y's image in the registry. What am I missing here? Following my reasoning the check for access would be made not during the ImageStreamTag pull but during its creation instead.

/cc @dmage

smarterclayton: Can you specifically describe the stream spec you expect?

```
spec:
  tags:
  - name: foo
    ref:
      name: bar:1
      namespace: baz
      kind: ImageStreamTag
```

?  Taking that reference automatically resolves it to a SHA, so if you want to get at the ref yes you can chase it BUT you can only chase up to the point that a) it becomes something other than an ImageStreamTag/Image AND b) the SHA in the ref is the same.  

So when you go from image stream 1 to image stream 2, you have to find the status tag with the same digest and use that ref (working backwards from newest).  If you follow those rules, you don't have to check ACL, but you MUST enforce those conditions to be safe.

I'd like to see an example set of yaml so I can make sure we're on the same page.

ricardomaraschini: For context: what I am trying to address is this [BZ](https://bugzilla.redhat.com/show_bug.cgi?id=1843253). In it the reference from one image stream to another is made using `DockerImage`, no `ImageStream{Tag,Image}` reference is involved. To reproduce this behavior I ran two commands:

```
$ oc tag -n openshift docker.io/library/centos:7 os:centos
Tag os:centos set to docker.io/library/centos:7.
$ oc tag -n default image-registry.openshift-image-registry.svc:5000/openshift/os:centos os:centos
Tag os:centos set to image-registry.openshift-image-registry.svc:5000/openshift/os:centos.
$
```

The (redacted) outcome in `openshift` namespace looks like this:

```yaml
kind: ImageStream
metadata:
  name: os
  namespace: openshift
spec:
  tags:
  - from:
      kind: DockerImage
      name: docker.io/library/centos:7
    name: centos
status:
  tags:
  - items:
    - created: "2021-04-26T09:11:32Z"
      dockerImageReference: docker.io/library/centos@sha256:e4ca2ed0202e76be184e75fb26d14bf974193579039d5573fb2348664deef76e
      image: sha256:e4ca2ed0202e76be184e75fb26d14bf974193579039d5573fb2348664deef76e
    tag: centos
```

Meanwhile in the `default` namespace we have the following image stream (redacted):

```yaml
metadata:
  name: os
  namespace: default
spec:
  tags:
  - from:
      kind: DockerImage
      name: image-registry.openshift-image-registry.svc:5000/openshift/os:centos
    name: centos
status:
  dockerImageRepository: image-registry.openshift-image-registry.svc:5000/default/os
  publicDockerImageRepository: default-route-openshift-image-registry.apps-crc.testing/default/os
  tags:
  - items:
    - created: "2021-04-26T09:12:18Z"
      dockerImageReference: image-registry.openshift-image-registry.svc:5000/openshift/os@sha256:e4ca2ed0202e76be184e75fb26d14bf974193579039d5573fb2348664deef76e
      image: sha256:e4ca2ed0202e76be184e75fb26d14bf974193579039d5573fb2348664deef76e
    tag: centos
```

At this point (assuming `openshift/os:centos` has not yet been pulled through) an attempt to pull from `default/os:centos` fails. My assumption was that if the image stream in `default` namespace got correctly created there are enough permissions in place to access `openshift/os:centos` from `default` namespace. What do you reckon?

stevekuznetsov:
@smarterclayton @ricardomaraschini any updates on this? We're actively having to work around this in our CI cluster suite.

dmage:
/hold

There change is getting complicated, so we'll need it to be covered by tests.

pkg/dockerregistry/server/pullthroughmanifestservice.go
dmage: GetImageOfImageStream is an API call, it doesn't pull the image. So I guess it doesn't do what you want it to do.

dmage:
@stevekuznetsov we are not as close to a solution as you might think

wewang58:
@ricardomaraschini Could you help to rebase the pr, so I can use cluster-bot to launch a cluster, thanks

PR #274, opened 2021-05-12
Bug 1960619: Use sha256 oauth tokens in tests

dmage:
The current method is deprecated. The tests should use sha256 tokens that are recommended to use.


dmage:
/retitle Bug 1960619: Use sha256 oauth tokens in tests

ricardomaraschini (approved):
/lgtm

PR #278, opened 2021-06-07
Bug 1918376: Bumping openshift/library-go

ricardomaraschini:
Bumping library-go to its last (as of today) master version.


dmage:
/override ci/prow/e2e-aws-upgrade

the networking problem is not related to the registry

PR #281, opened 2021-06-17
Credentials store factory

ricardomaraschini:


xiuwang:
Verified on disconnect cluster.
Could pull image from the mirror registry with auth via imagecontentsourcepolicy

/label qe-approved

PR #295, opened 2021-09-15
IR-104: Use library-go in image-registry

Nikita757:
Replaced some of the origin-common code with more supported library-go package.


dmage:
@Nikita757 this PR doesn't delete any function/package from this repository, can't we delete something?

dmage:
@Nikita757 please update README files for kubernetes-common and origin-common, and rebase it on top of master so that you don't have a merge commit in the PR.

pkg/dockerregistry/server/quotarestrictedblobstore.go
```diff
@@ -153,7 +153,8 @@ func admitBlobWrite(ctx context.Context, repo *repository, size int64) error {
 
 // admitImage checks if the size is greater than the limit range.
 func admitImage(size int64, limit corev1.LimitRangeItem) error {
-	if limit.Type != imageapi.LimitTypeImage {
+	if limit.Type !=
+		imageapiv1.LimitTypeImage {
```
dmage, line 157: Why on a different line?

pkg/dockerregistry/server/prune/prune.go
```diff
@@ -229,7 +229,7 @@ func Prune(ctx context.Context, registry distribution.Namespace, registryClient
 			return fmt.Errorf("failed to parse the repo name %s: %v", repoName, err)
 		}
 
-		ref, err := imageapi.ParseDockerImageReference(repoName)
+		ref, err := imageref.Parse(repoName)
```
dmage, line 232: Should it be imageutil.ParseDockerImageReference?

pkg/dockerregistry/server/quotarestrictedblobstore.go
Nikita757: A typo

pkg/dockerregistry/server/prune/prune.go
Nikita757: Indeed imageref.Parse() is a better choice. I'll replace them.

xiuwang:
Do regression for image-registry, no issue found.
/lgtm

PR #319, opened 2022-03-18
fake bump to get revert

deads2k:


deads2k:
/hold

this is just to test our options with clusterbot.

deads2k:
Test from https://github.com/openshift/origin/pull/26920 demonstrates that this allows redirects to function again.

`make && openshift-tests run-test "[sig-imageregistry] Image registry should redirect on blob pull [Suite:openshift/conformance/parallel]"`

deads2k:
now in https://github.com/openshift/image-registry/pull/320

PR #325, opened 2022-04-25
IR-253: report push and pull metrics

flavianmissi:
Push and pull metrics are reported for manifest GET and PUT.

image


flavianmissi:
we discussed this with @dmage and it turns out that there are already metrics for these operations, we just need to write some prometheus rules to export only some specific fields from those.
useful queries:
```
max by (operation) (imageregistry_request_duration_seconds_count)


imageregistry_request_duration_seconds_count{operation=~"ManifestService\\..*"}
```

closing.

PR #329, opened 2022-05-10
Bug 1923536: forward http.StatusTooManyRequests to client

flavianmissi:
makes the registry forward http.StatusTooManyRequests back
to the client during pullthrough operations.


dmage:
I'm ok with keeping these debug logs, they are disabled by default anyway.
/approve

xiuwang:
/label qe-approved
Add verify steps in bugzilla 1923536

dmage:
Ideally we should have new tests in every bug fix, but given this is a very long standing issue, let's merge it!
/lgtm

flavianmissi:
> Ideally we should have new tests in every bug fix, but given this is a very long standing issue, let's merge it!

agreed! can still write the tests, but I'll need some help - spent a few hours when I was writing this but couldn't really grasp how I could cover it :/

PR #331, opened 2022-05-13
Bug 2067893: apply upstream prometheus/client_golang fix

flavianmissi:
Taken from https://github.com/prometheus/client_golang/pull/987
and manually applied to relevant files.


dmage:
can you use `glide` to get new client_golang?

flavianmissi:
> can you use `glide` to get new client_golang?

It doesn't work. Glide tries to update all packages (there seems to be no way to update a single pkg).
Which fails:
```
[ERROR]	Error scanning github.com/cespare/xxhash/v2: cannot find package "." in:
	~/.glide/cache/src/https-github.com-cespare-xxhash/v2
```
`github.com/cespare/xxhash` doesn't have a `v2` directory so we can't install it without go mod.

PR #337, opened 2022-06-28
on server shutdown use the termination grace period of 55 seconds

rphillips:


pkg/cmd/dockerregistry/dockerregistry.go
```diff
@@ -188,9 +188,8 @@ func Execute(configFile io.Reader) {
 	}
 
 	// signal has been received. we will attempt to stop our http server
-	// using a timeout slightly lower than the default k8s grace period
-	// of 30 seconds.
-	ctx, cancel := context.WithTimeout(ctx, 29*time.Second)
+	// using a timeout equal to the terminationGracePeriod.
+	ctx, cancel := context.WithTimeout(ctx, 55*time.Second)
```
bparees, line 192: my understanding is that the motivation for this change is to ensure that the readiness probe has failed before the pod fully shuts down.  (the readiness probe is 3 failures at 10s interval, so it'll take 30s for ready to go false, leaving no buffer before the pod disappears after going unready)

that said, I think even *better* behavior would be for the readiness endpoint to stop serving traffic, but for the rest of the pod to continue accepting connections, at least for some period of time.  That would ensure that the readiness=false has propagated and things stop sending it new traffic, before it stops accepting new traffic.

i.e.:

1) termination requested
2) stop responding to readiness probes
3) after ~30s we know ready=false so it should now be safe to stop accepting all new traffic (modulo some buffer for readiness + associated endpointslice updates to propagate to loadbalancers/routers)
4) *now* stop accepting new connections while continuing to drain existing connections
5) some time later either shutdown because remaining connections terminated, or give up after 20s and terminate.

bparees: (such a change would also require that the liveness probe and readiness probe stop using the same endpoint, otherwise the liveness probe failure could prematurely kill the pod)

dmage:
/hold
@rphillips can you share details about these PRs? This grace period affects only long standing pushes, and we have it just to avoid warnings in logs. The longer period here won't change anything, but will cause noise in logs.

pkg/cmd/dockerregistry/dockerregistry.go
dmage: > that said, I think even better behavior would be for the readiness endpoint to stop serving traffic

afaik that's exactly what happens here - https://github.com/openshift/image-registry/blob/9f07f4370075b3a820f3517ca5a9da339404ba9b/pkg/cmd/dockerregistry/dockerregistry.go#L196

That's how it looks like today:

1. Termination requested
2. Kubernetes initiates removing the pod from services and routes.
3. We hope that in 25 seconds the pod will be removed from its services, so we have a pre-stop action `sleep 25` (you can find the same for the console authored by Clayton).
4. Once we don't expect new incoming connections, the registry can reject new incoming connections. So once the pre-stop action is complete, the registry gets SIGTERM and shutdowns the listening socket.
5. It waits for 29 seconds until clients finish their in-progress requests.
6. If there are an active connection, it'll be terminated with the registry. Otherwise the registry can finish its job once all clients are gone.

rphillips:
> Kubernetes initiates removing the pod from services and routes.

1.22 introduced changes where terminating pods are not removed from load balancers/services until the pod is either terminated or the readiness probe goes to false during termination.

We don't want this PR... We will want my alternative PR to make the readiness probes return false during shutdown.

https://github.com/openshift/image-registry/pull/338

/close

dmage:
How other applications should be adapted to such a change? I doubt there are applications that have an endpoint that says "we are going to stop accepting new connections in few seconds"

pkg/cmd/dockerregistry/dockerregistry.go
bparees: > That's how it looks like today:

it's close, but not exactly the same because it does not result in readiness=false report prior to the pod stopping accepting new connections.

Which isn't supposed to matter since the pod is terminating and thus should be removed from endpoints, but as Ryan noted, that logic/assumption is changing in k8s, where pods can remain in the endpointslice but with a ready=false condition, instead of being removed immediately.

PR #338, opened 2022-07-01
WIP: create a readyz endpoint

rphillips:


pkg/cmd/dockerregistry/dockerregistry.go
```diff
@@ -187,18 +189,25 @@ func Execute(configFile io.Reader) {
...
 	}
 
-	// signal has been received. we will attempt to stop our http server
-	// using a timeout slightly lower than the default k8s grace period
-	// of 30 seconds.
-	ctx, cancel := context.WithTimeout(ctx, 29*time.Second)
-	defer cancel()
+	dcontext.GetLogger(ctx).Infof("setting service to unready")
+	setReady(false)
+	time.Sleep(35 * time.Second)
```
bparees, line 194: we still want this to be 55s right?  i.e. we want to stick around as long as possible closing existing connections, right up until the termination grace period(which is 55s) expires if necessary

rphillips: We want this to be readinessPeriod*Failurethreshold (10x3+fudge=35)... this allows the readiness to get propagated to the api/load-balancers/etc. After this... The rest of the shutdown proceeds. I removed the timeout on the HTTP server shutdown. By removing the http server timeout all the connections will gracefully terminate. The pod has a hard stop of 55 seconds already with the terminationGracePeriodSeconds being set to 55.

dmage:
How other applications should be adapted to such a change? I doubt there are applications that have an endpoint that says "we are going to stop accepting new connections in few seconds". Ideally you should assume that the registry is just yet another external application (as we have Quay that should haven't have disruptions too).

bparees:
> How other applications should be adapted to such a change? I doubt there are applications that have an endpoint that says "we are going to stop accepting new connections in few seconds".

it does seem like it would be nice if k8s could manage all of this for the app (i.e. the app could opt into saying "when my pod is about to be terminated, first mark this pod's endpoint as ready=false, then after X seconds, start the termination process for the pod").  This wasn't as important when starting the termination process for a pod explicitly removed it from endpoints, but the new behavior makes it harder for apps to get this right.

Such a an option would remove the need for every application pod to implement (some of) this multistep shutdown logic itself.

@rphillips does that seem potentially worth an upstream KEP?

rphillips:
I'm not sure about the history of the change. Perhaps @deads2k can shed more light on it.

PR #346, opened 2022-09-22
[OCPNODE-1258] Support both icsp and idms

QiWang19:
Append ImageDigestMirrorSet to current ImageContentSourcePolicy. The [OCPNODE-521](https://issues.redhat.com//browse/OCPNODE-521) will land new CRDs for image mirror configurations to replace ICSP in the future release.

ImageDigestMirrorSet implemented: https://github.com/openshift/machine-config-operator/pull/3037
ImageDigestMirrorSet epic: https://issues.redhat.com/browse/OCPNODE-521


QiWang19:
/hold
for 4.13 tree to open

QiWang19:
/hold cancel
4.13 branch is open

QiWang19:
@dmage 4.13 branch is open, this is ready for review. I have rebased this branch, not sure why the ci failed. Could you help take a look?

QiWang19:
failure TestPullThroughIDMS: https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_image-registry/346/pull-ci-openshift-image-registry-master-e2e-agnostic-image-registry/1612831060893110272
The test might reply on the PR: https://github.com/openshift/machine-config-operator/pull/3037 that hasn't been merged yet.

pkg/dockerregistry/server/simplelookupicsp.go
```diff
@@ -41,7 +47,18 @@ func (s *simpleLookupICSP) FirstRequest(
...
 
-	imageRefList, err := s.alternativeImageSources(ref, icspList.Items)
+	idmsList, err := s.idmsClient.List(ctx, metav1.ListOptions{})
+	if err != nil {
+		klog.Errorf("unable to list IDMS config: %s", err)
+		return []reference.DockerImageReference{ref.AsRepository()}, nil
+	}
+
+	if len(icspList.Items) > 0 && len(idmsList.Items) > 0 {
+		klog.Errorf("error finding both icsp and idms resources at the same time: %s", err)
```
flavianmissi, line 57: Can we clarify this message a bit? It caused me some confusion the way it's currently phrased.
What about:
```
error: found both ICSP and IDMS resources, but only one or the other is supported
```
Additionally, you're adding the error itself to the formatting, but at this point the error will be nil.

pkg/dockerregistry/server/simplelookupicsp.go
```diff
@@ -41,7 +47,18 @@ func (s *simpleLookupICSP) FirstRequest(
...
+		return []reference.DockerImageReference{ref.AsRepository()}, nil
```
flavianmissi, line 58: I think we should surface an error here so that this doesn't go unnoticed to the caller.

pkg/dockerregistry/server/simplelookupicsp_test.go
```diff
@@ -337,29 +284,56 @@ func TestFirstRequest(t *testing.T) {
...
+			if err != nil {
+				t.Fatalf("FirstRequest does not return error, received: %s", err)
+			}
+
+			if !reflect.DeepEqual(alternates, tt.res) {
+				t.Errorf("expected %+v, received %+v", tt.res, alternates)
+			}
+		})
+
+		t.Run(tt.name, func(t *testing.T) {
```
flavianmissi, line 327: Can you append something to the test case name to indicate that this sub-test is for IDSM rules?
Same goes for the ICSP rules sub-test above on line 300.

pkg/dockerregistry/server/simplelookupicsp.go
```diff
@@ -17,15 +20,18 @@ import (
 // *Context AlternativeBlobSourceStrategy interface.
 type simpleLookupICSP struct {
 	icspClient operatorv1alpha1client.ImageContentSourcePolicyInterface
+	idmsClient cfgv1client.ImageDigestMirrorSetInterface
 }
 
 // NewSimpleLookupICSPStrategy returns a new entity of simpleLookupICSP using provided client
 // to obtain cluster wide ICSP configuration.
 func NewSimpleLookupICSPStrategy(
```
flavianmissi, line 28: Do you have a plan for when to rename this function (and the related types)?

pkg/dockerregistry/server/simplelookupicsp.go
```diff
@@ -17,15 +20,18 @@ import (
 // *Context AlternativeBlobSourceStrategy interface.
 type simpleLookupICSP struct {
 	icspClient operatorv1alpha1client.ImageContentSourcePolicyInterface
+	idmsClient cfgv1client.ImageDigestMirrorSetInterface
 }
 
 // NewSimpleLookupICSPStrategy returns a new entity of simpleLookupICSP using provided client
 // to obtain cluster wide ICSP configuration.
```
flavianmissi, line 27: Can you update this function's comment so that it reflects the fact that the configuration may also come from the IDSM?

flavianmissi (commented):
Looks pretty good, thanks for doing this.
Left a few comments, please take a look.

test/integration/pullthrough/pullthrough_test.go
```diff
@@ -423,3 +425,178 @@ func TestPullThroughICSP(t *testing.T) {
...
+							cfgv1.ImageMirror(remoteRegistryAddr + "/" + isname),
+						},
+					},
+				},
+			},
+		},
+		metav1.CreateOptions{},
+	)
+	if err != nil {
+		t.Fatalf("error creating ICSP rule: %s", err)
```
flavianmissi, line 493: `s/ICSP/IDMS`

flavianmissi:
@QiWang19 I'm not sure about your failing test. You can try running it against your own cluster with
``
`KUBECONFIG=<path-to-kubeconfig> GOTEST_FLAGS="-p 1 -mod vendor" DOCKER_API_VERSION=1.24 hack/test-go.sh test/integration/*
```

**Comment by QiWang19** [comment]
> @QiWang19 I'm not sure about your failing test. You can try running it against your own cluster with
> 
> ```
> KUBECONFIG=<path-to-kubeconfig> GOTEST_FLAGS="-p 1 -mod vendor" DOCKER_API_VERSION=1.24 hack/test-go.sh test/integration/*
> ```

Thanks! I will try running the test to make sure the code in this patch is correct.

**Comment by QiWang19** [comment]
The error is 
```
--- FAIL: TestPullThroughIDMS (9.79s)
    project.go:42: project image-registry-test-integration-idms is created
    registry.go:505: created ephemeral registry: ephemeral-registry-jznd4-image-registry-test-integration-idms.apps.ci-ln-6g0lzsk-76ef8.origin-ci-int-aws.dev.rhcloud.com (ephemeral-registry-jznd4)
    pullthrough_test.go:542: unexpected status 0: v1.Status{TypeMeta:v1.TypeMeta{Kind:"", APIVersion:""}, ListMeta:v1.ListMeta{SelfLink:"", ResourceVersion:"", Continue:"", RemainingItemCount:(*int64)(nil)}, Status:"Failure", Message:"Internal error occurred: does.not.exist/repo/image@sha256:c1d7a836008fcd9cd6cd2b4b4735bd0c2fef0590b468aedd84abbd6d476032e0: Get \"http://does.not.exist/v2/\": dial tcp: lookup does.not.exist on 172.30.0.10:53: no such host", Reason:"InternalError", Details:(*v1.StatusDetails)(0xc000cb3c80), Code:500}


```
But this test reuses most of `TestPullThroughICSP`. I need to look into why it failed.

**Comment by dmage** [comment]
@QiWang19 the error seems to be from image streams, did you implement IDMS for image streams? (they live inside openshift/openshift-apiserver)

**Comment by QiWang19** [comment]
Thanks. @dmage @flavianmissi 
no I did not implement openshift-apiserver. I found here: https://github.com/openshift/openshift-apiserver/blob/94da954fae6d5512de5073e3f30415fcec44152e/pkg/image/apiserver/registry/imagestreamimport/rest.go#L283 will work on t idms.

**Comment by flavianmissi** [comment]
@QiWang19 I suspect that pull-through test won't pass until https://github.com/openshift/openshift-apiserver/pull/318 is merged... You might want to validate that by deploying a cluster with this PR and https://github.com/openshift/openshift-apiserver/pull/318 together, then running your test again on that cluster. If the test passes then we just need to wait, otherwise you have an actual test failure to troubleshoot.

Relatedly, I see that https://github.com/openshift/openshift-apiserver/pull/318 adds support for both `ImageDigestMirrorSet` and `ImageTagMirrorSet`, but here you only covered `ImageDigestMirrorSet`. Do we not do pull-through by tag? 🤔

**Comment by QiWang19** [comment]
The PR was failing to launch a 4.13 cluster, I will continue to retest the integration test.



> Relatedly, I see that [openshift/openshift-apiserver#318](https://github.com/openshift/openshift-apiserver/pull/318) adds support for both `ImageDigestMirrorSet` and `ImageTagMirrorSet`, but here you only covered `ImageDigestMirrorSet`. Do we not do pull-through by tag? thinking

@flavianmissi  Current image-registry only looks up icsp objects that supports only the digest image pull. I didn't know if the image-registry want to add the `imageTagMirrorSet` for tag pulls, so only added ImageDigestMirrorSet to keep the same image pull behavior as with ICSP.  I can update the PR to add support for tag pulls. What do you think? 
It seems the openshift-apiserver and image-registry should be in alignment.

**Comment by flavianmissi** [comment]
yeah I do think it makes sense to align with the openshft-apiserver. we want to avoid any confusion, and having different behaviour would likely not help.

**Comment by QiWang19** [comment]
Tested this PR with [openshift-apiserver/pull/318](https://github.com/openshift/openshift-apiserver/pull/318) on ovirt cluster set up by Cluster Bot, tests passed
```
launch https://github.com/openshift/openshift-apiserver/pull/318 ovirt

$ KUBECONFIG=/home/qiwan/kubeletconfig GOTEST_FLAGS="-p 1 -mod vendor" DOCKER_API_VERSION=1.24 hack/test-go.sh test/integration/*
[INFO] [CLEANUP] Cleaning up temporary directories
ok  	github.com/openshift/image-registry/test/integration/crossmount	(cached)	coverage: [no statements]
ok  	github.com/openshift/image-registry/test/integration/deletemanifest	(cached)	coverage: [no statements]
ok  	github.com/openshift/image-registry/test/integration/imagelayers	(cached)	coverage: [no statements]
ok  	github.com/openshift/image-registry/test/integration/metricspullthrough	(cached)	coverage: [no statements]
ok  	github.com/openshift/image-registry/test/integration/oci	(cached)	coverage: [no statements]
ok  	github.com/openshift/image-registry/test/integration/offline	(cached)	coverage: [no statements]
ok  	github.com/openshift/image-registry/test/integration/pullthrough	57.900s	coverage: [no statements]
ok  	github.com/openshift/image-registry/test/integration/pullthroughblob	(cached)	coverage: [no statements]
ok  	github.com/openshift/image-registry/test/integration/requestloop	(cached)	coverage: [no statements]
ok  	github.com/openshift/image-registry/test/integration/uploadblob	(cached)	coverage: [no statements]
ok  	github.com/openshift/image-registry/test/integration/v2	(cached)	coverage: [no statements]
[INFO] hack/test-go.sh exited with code 0 after 00h 01m 01s
```

flavianmissi:
Hi @QiWang19,

> in the previous ci test cluster 4.13.0-0.ci, there was a default imagecontentsoucepolicy `image-policy-0`

hmmm are we landing any default IDMS/ITMS? in other words, will customers face what you're seeing? I'm thinking that as long as the there aren't IDMS/ITMS defaults deployed WITH ICSP defaults then everything should work, but I don't about defaults.

additionally (and unrelated), the registry does not need ITMS because pull-through is always made by digest. BUT it's best to stay in line with other implementations (such as the one in openshift-apiserver), so let's keep both. There's just no way of testing ITMS via the registry alone, since it won't pull-through by tag.

flavianmissi:
Based on this:

> Having a default ICSP on the cluster can cause ci test failure to migration PRs because using both CRDs for mirror setting is prohibited. Will also need to figure out the migration of the CI cluster.

I'm wondering if it's easier to allow ICSP to exist along with IDMS, but prefer IDMS when present (maybe log some warnings when this happens?). Then in 4.14 we can add back the code to disallow them to co-exist. It would make migration easier.
What do you think @QiWang19 ?

QiWang19:
Hi @flavianmissi, it has been discussed that IDSM and ICSP should not on the same cluster at the start, since they configure the same underlying file registries.conf. We can avoid different clients update the same configuration. We prevent this from the admission https://github.com/openshift/kubernetes/blob/master/openshift-kube-apiserver/admission/customresourcevalidation/imagedigestmirrorset/validate_imagedigestmirrorset.go
I'd like to figure out other ways so the discussion does not need to go back and forth.

flavianmissi:
That sounds like a plan @QiWang19 .

The migration of the CI cluster to use IDMS by default instead of ICSP will need to happen before we merge this PR, right? Or do I misunderstand?

Additionally, are we planning to drop ICSP support in 4.14? We should only remove ICSP related tests when we remove the feature entirely, code and all. Are these tests hindering us in some way?

QiWang19:
> Additionally, are we planning to drop ICSP support in 4.14?

The ICSP will be dropped at least after 4.17 to allow the EUS to EUS upgrade.
Yes, CI cluster migration will need to happen before we merge this PR.

> We should only remove ICSP related tests when we remove the feature entirely

Given that not all CI clusters use default IDMS/ICSP,  one way to keep the ICSP test by 4.17 is before the `TestPullThroughICSP` function creates ICSP, in this PR we add a check to see if there's a default IDMS on the cluster, if true, we return with `t.Log("default IDMS mirror configuration exist, can't create ImageContentSourcePolicy when  ImageDigestMirrorSet resources exist. Consider update the test)`. So we can avoid the admission error.

flavianmissi:
> The ICSP will be dropped at least after 4.17 to allow the EUS to EUS upgrade.

that makes sense.

> Given that not all CI clusters use default IDMS/ICSP, one way to keep the ICSP test by 4.17 is before the TestPullThroughICSP function creates ICSP, in this PR we add a check to see if there's a default IDMS on the cluster [...]

that seems to be the best course of action for now. can you add it to this PR then?

flavianmissi:
I think master is open to 4.14 development now @QiWang19 🎉

pkg/dockerregistry/server/simplelookupicsp.go
QiWang19: Addressed this in line68 and line73.

pkg/dockerregistry/server/simplelookupicsp.go
QiWang19: Updated to rename the field in this PR.

QiWang19:
@flavianmissi Thanks for letting me know about the development for the next release. :heart: 
I have rebased this PR and addressed the previous reviews. This is ready for another round of review.
We will hold this before the ci cluster gets updated with the IDMS object per our discussion above.

QiWang19:
@flavianmissi I rebased https://github.com/openshift/openshift-apiserver/pull/318, do you know which part of the ci script is related to this PR for the pre-merge test?
I have a draft PR: https://github.com/openshift/release/pull/39122

flavianmissi:
> do you know which part of the ci script is related to this PR for the pre-merge test?
@QiWang19 do you mean where are the pre-merge tests for image stream import located? I'm actually not sure 🤔 
@dmage do you happen to know? I looked around a bit but it wasn't obvious (to me...)

dmage:
@QiWang19 what job are you talking about? openshift/release has a special logic with ICSP for "disconnected" clusters, which is not specific to the image-registry repo nor openshift-apiserver.

dmage:
@QiWang19 what job are you talking about? openshift/release has a special logic with ICSP for "disconnected" clusters, which is not specific to the image-registry repo nor openshift-apiserver.

QiWang19:
@dmage This PR failed the test cases creating IDMS objects because the ci cluster has an existing ICSP objects. I am thinking about updating the openshft/release script can change the existing ICSP object to IDMS object so this PR can pass the test. 
But I am not sure if changing openshift/release logic from ICSP to IDMS requires IDMS support from this image-registry

> which is not specific to the image-registry repo nor openshift-apiserver

So the openshift/release logic with ICSP does not rely on image-registry?

dmage:
@QiWang19 `openshift/release` is a huge thing, it runs on OCP and uses additional OCP clusters to install and test other OCP clusters; all these clusters have a few instances of the image registry.

The latest test results for this PR are few months old and don't seem to be related to this PR.
/retest

QiWang19:
@dmage  I just remembered your [comment](https://github.com/openshift/image-registry/pull/346#issuecomment-1379378032) we should https://github.com/openshift/openshift-apiserver/pull/318 should get this in first.

Can we merge that or that PR still needs to hold for pre-merge testing? I rebased that 3 weeks ago.

dinhxuanvu:
@QiWang19 FYI, the rebase is merged.

QiWang19:
@flavianmissi I rebased after https://github.com/openshift/openshift-apiserver/pull/318 merged. Can we get this in?

QiWang19:
@dmage Can we get this in?

PR #347, opened 2022-09-26
IR-269: Enable pull-through for manifest lists

dmage:


pkg/dockerregistry/server/pullthroughmanifestservice.go
```diff
@@ -51,27 +51,39 @@ func (m *pullthroughManifestService) Get(ctx context.Context, dgst digest.Digest
...
+			return nil, rErr
 		}
-		return nil, rErr
-	}
-
-	ref, err := reference.Parse(image.DockerImageReference)
-	if err != nil {
-		dcontext.GetLogger(ctx).Errorf("bad DockerImageReference (%q) in Image %s@%s: %v", image.DockerImageReference, m.imageStream.Reference(), dgst.String(), err)
-		return nil, err
+	} else {
```
flavianmissi, line 78: why the `else`? it's already confusing to do a `if a != b {} else {}` (instead of `if a == b {} else {}`).
and in this case the `swtich` inside the `if` block has a default which returns, so when `if rErr != nil` is true won't the function always return, thus making the `else` unnecessary?

xiuwang:
Reported a bug https://issues.redhat.com/browse/OCPBUGS-4282

pkg/dockerregistry/server/pullthroughmanifestservice.go
flavianmissi: nevermind, I realise that not every `case` inside the previous switch returns so we do need the `else` here

pkg/imagestream/imagestream.go
```diff
@@ -189,6 +190,62 @@ func (is *imageStream) GetImageOfImageStream(ctx context.Context, dgst digest.Di
...
+		if found {
+			parent = image
+			break
+		}
+	}
+	if parent == "" {
+		return reference.DockerImageReference{}, rerrors.NewError(
+			ErrImageStreamImageNotFoundCode,
+			fmt.Sprintf("ResolveUpstreamRef: unable to find parent for image %s in image stream %s", dgst.String(), is.Reference()),
+			nil,
```
flavianmissi, line 221: This can't be nil, you can give it a `kerrors.NewNotFound(imageapiv1.Resource("imagestreamimage"), dgst.String())`. It'll weirdly panic and recover otherwise - at least that's what I've seen on unit tests.

flavianmissi: This is what I've seen in the tests:
```
    logrus.go:29: time="2022-12-23T14:56:11+01:00" level=error msg="manifestService.Get: unable to get image sha256:abd132 in imagestream user/app: %!v(PANIC=Error method: runtime error: invalid memory address or nil pointer dereference)"
```
It was probably me causing it by refactoring the openshifit fake, but still.
(I temporarily stole ResolveUpstreamRef)

Prashanth684:
from: https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/pr-logs/pull/openshift_image-registry/347/pull-ci-openshift-image-registry-master-e2e-agnostic-image-registry/1613913706297561088/artifacts/e2e-agnostic-image-registry/e2e-test/artifacts/unit.log , looks like it's failing with: 

```
{"Time":"2023-01-13T16:15:53.435432239Z","Action":"output","Package":"github.com/openshift/image-registry/test/integration/pullthrough","Test":"TestPullThroughInsecure","Output":"    pullthrough_test.go:273: unexpexted access to insecure blobs\n"}
```

flavianmissi:
thanks Oleg!
/lgtm

flavianmissi:
/assign @stevsmit
/assign @sferich888
for docs and px approvals

PR #355, opened 2022-11-25
IR-270: support push by digest

flavianmissi:


flavianmissi:
Unit tests will not pass until https://github.com/openshift/image-registry/pull/347 lands. Code should be ready for review otherwise
/hold 
/assign @dmage

flavianmissi:
/hold
I'll look into making the necessary changes in openshift-apiserver instead. Holding this until I know which approach to take.

PR #359, opened 2023-01-27
IR-341: bump openshift/api

flavianmissi:
this time taking care to consistently use the same k8s version


flavianmissi:
This is an internal change only.
/label px-approved
/label docs-approved

PR #376, opened 2023-08-30
OCPBUGS-18321: bump docker-distribution

flavianmissi:


flavianmissi:
need to update the reference to openshift/docker-distribution after https://github.com/openshift/docker-distribution/pull/37 merges
/hold

flavianmissi:
updated reference to openshift/docker-distribution instead of my own fork.
/hold cancel

PR #377, opened 2023-09-04
Allow ICSP IDMS coexisting

QiWang19:
Allow ICSP IDMS coexisting


QiWang19:
@flavianmissi could you review?

PR #386, opened 2023-11-21
Add support for il-central-1 region in AWS

sadasu:


flavianmissi:
this is an internal change and doesn't affect users directly.
/label docs-approved
/label px-approved

flavianmissi:
@wewang58 could you help with qe-approval? Regression tests should be good enough.

PR #392, opened 2024-02-08
OCPBUGS-29003: update distribution with bugfix

flavianmissi:


flavianmissi:
The `e2e-metal-ipi-ovn-ipv6` has failed due to different reasons each time, all unrelated to the changes introduced by this PR. I think it's safe to say the failures are unrelated.
/override ci/prow/e2e-metal-ipi-ovn-ipv6

PR #393, opened 2024-02-16
[release-4.15] OCPBUGS-29525: bump docker-distribution to fix azure path issue

flavianmissi:


sdodson:
This is necessary for an urgent 4.14 fix, this is not critical for 4.15 GA.

PR #402, opened 2024-06-04
OCPBUGS-33453: Use SAR instead of SSAR

ibihim:
## What

Use `subjectaccessreviews` instead of `selfsubjectaccessreviews` for access denied and check for anonymous access.

## Why

SSAR doesn't work anymore with [AUTH-509](https://issues.redhat.com/browse/AUTH-509), but we can do a SAR instead and specifying the user / group as anonymous to avoid a regression in 4.16

## Note

- Depends on https://github.com/openshift/cluster-image-registry-operator/pull/1046
- WIP, need to clean it up


flavianmissi:
Looks good to me, but please have a look at QE's latest comment on the bug: https://issues.redhat.com/browse/OCPBUGS-33453?focusedId=24860861&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-24860861 seems like something is off.

PR #418, opened 2025-01-15
OCPBUGS-38874: add ap-southeast-5 region

ricardomaraschini:
this pr adds the ap-southeast-5 region to the list of supported regions. this change is made directly into the vendor directory and is intended only for test purposes. if we get the ok that this works we will make the changes and bumps necessary to land this in the right place.

### DIFF
--- a/vendor/github.com/distribution/distribution/v3/registry/storage/driver/s3-aws/s3.go
+++ b/vendor/github.com/distribution/distribution/v3/registry/storage/driver/s3-aws/s3.go
@@ -42,6 +42,13 @@ import (
 	"github.com/distribution/distribution/v3/registry/storage/driver/factory"
 )
 
+// additionalRegions is a slice with regions not listed among the default
+// regions supported by aws-sdk-go v1. the v1 of the sdk is in maintenance
+// therefore it does not receive new regions as aws infrastructure grows.
+var additionalRegions = []string{
+	"ap-southeast-5",
+}
+
 const driverName = "s3aws"
 
 // minChunkSize defines the minimum multipart upload chunk size
@@ -128,6 +135,10 @@ func init() {
 		for region := range p.Regions() {
 			validRegions[region] = struct{}{}
 		}
+	}
+
+	for _, region := range additionalRegions {
+		validRegions[region] = struct{}{}
 	}
 
 	for _, objectACL := range []string{


flavianmissi:
Inspection of `e2e-aws-image-registry` [failed jobs](https://prow.ci.openshift.org/pr-history/?org=openshift&repo=image-registry&pr=418) show a variation of the following failure over and over again:
```
[sig-imageregistry][Serial] Image signature workflow can push a signed image to openshift registry and verify it [apigroup:user.openshift.io][apigroup:image.openshift.io] [Skipped:Disconnected] [Suite:openshift/conformance/serial]
{  fail [github.com/openshift/origin/test/extended/images/signatures.go:40]: Unexpected error:
    <*errors.errorString | 0xc001a1c270>: 
    timed out while waiting of an image stream tag e2e-test-registry-signing-kvgq5/signer:latest
    {
        s: "timed out while waiting of an image stream tag e2e-test-registry-signing-kvgq5/signer:latest",
    }
occurred
Ginkgo exit error 1: exit with code 1}
```

This indicates the failure is probably not a flake, though it's not necessarily related to this change.

We can use https://search.ci.openshift.org/ to search for job results in all CI jobs to see if the same test fails in jobs unrelated to this PR:
https://search.dptools.openshift.org/?search=Image+signature+workflow+can+push+a+signed+image+to+openshift+registry+and+verify+it&maxAge=48h&context=1&type=bug%2Bissue%2Bjunit&name=&excludeName=&maxMatches=5&maxBytes=20971520&groupBy=job

There we can see that there are several pre-merge failures from different repositories. Look for `pull-ci-openshift-image-registry` for the list of job failures coming from this PR, the `openshift-image-registry` part matches the org/repo names in Github for every `pull-ci` entry, and the `pull-ci` part indicates pre-merge jobs - from this you can determine this failure is also affecting other repositories. This alone means we could probably override this test failure. I prefer to override tests only if I _must_.

ricardomaraschini:
> There we can see that there are several pre-merge failures from different repositories. Look for `pull-ci-openshift-image-registry` for the list of job failures coming from this PR, the `openshift-image-registry` part matches the org/repo names in Github for every `pull-ci` entry, and the `pull-ci` part indicates pre-merge jobs - from this you can determine this failure is also affecting other repositories. This alone means we could probably override this test failure. I prefer to override tests only if I _must_.

I looked into 3 of the failures you mentioned and they all fail due to the same reason: the test creates this [BuildConfig](https://github.com/openshift/origin/blob/master/test/extended/testdata/signer-buildconfig.yaml) and it fails with while building with:

```
error: build error: building at STEP "RUN source /etc/os-release     && rhel_major=${VERSION_ID%.*}     && yum config-manager     --add-repo "https://cdn-ubi.redhat.com/content/public/ubi/dist/ubi${rhel_major}/${rhel_major}/\$basearch/baseos/os/"     --add-repo "https://cdn-ubi.redhat.com/content/public/ubi/dist/ubi${rhel_major}/${rhel_major}/\$basearch/appstream/os/"": while running runtime: exit status 6
```

This is an [example](https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/test-platform-results/pr-logs/pull/29428/pull-ci-openshift-origin-master-e2e-aws-ovn-serial/1879630839508111360/artifacts/e2e-aws-ovn-serial/openshift-e2e-test/artifacts/junit/e2e-test-registry-signing-dpqk9/signer-1-build/docker-build/logs.txt) of what I think is the output of the build process.

adambkaplan:
Echoing my comment on Slack - this test probably was written before the UBI catalog existed. It looks like the skopeo image should be used as the base image, which allows you to skip all of the `yum/dnf` commands.

Using `registry.access.redhat.com/ubi9/skopeo:latest` should do the trick (or other tag if you want to pin on a specific version).

wewang58:
@ricardomaraschini  So the failed jobs are not related to the pr, right?

ricardomaraschini:
> @ricardomaraschini So the failed jobs are not related to the pr, right?

It seems like those are not related. I am trying to solve one of the tests failures at the moment.

ricardomaraschini:
This was a test PR, now we can safely close it. The actual fix is coming next.

PR #423, opened 2025-01-31
[release-4.16] OCPBUGS-49696: bump docker distribution

ricardomaraschini:
bumping docker-distribution dependency as the new version includes a few new regions.


flavianmissi:
/retitle [release-4.16] OCPBUGS-49696: bump docker distribution

PR #425, opened 2025-03-26
IR-522: disable aws region check

ricardomaraschini:
disable the artificial aws region check. we are doing this so we can support new aws regions without requiring patching the code every time.

this is a test and should not land on main.


wewang58:
Finished test, no critical issue found.

PR #437, opened 2025-06-30
IR-526: remove AWS region check

ricardomaraschini:
Bump docker-distribution to the last commit on `release-4.18` branch. This essentially disables the manual region check we used to do.


sdodson:
/override ci/prow/okd-scos-images
OKD is EOL on 4.18, these tests should be removed

PR #450, opened 2025-11-03
AGENT-1311: Add support for distribution/distribution registry binary

rwsu:
This PR adds support for running a generic distribution/distribution registry binary alongside the existing OpenShift-integrated registry (dockerregistry) in the same container image. This provides flexibility to run the image either as an OpenShift-integrated registry or as a standalone distribution/distribution registry.

Changes

  - Added cmd/distribution/main.go: New entry point for the generic registry binary that imports the upstream distribution/distribution registry code
  - Updated Dockerfile.ocp: Modified to copy both registry binaries (dockerregistry and distribution) and their respective configurations:
    - /config.yml for OpenShift-integrated registry 
    - /etc/registry/config.yaml for generic distribution/distribution registry
  - Added images/registry/config.yaml: Configuration file for the generic distribution/distribution registry sourced from
  https://github.com/openshift/appliance/blob/master/registry/config.yml, configured for filesystem storage at /var/lib/registry
  - Updated build configuration: Modified hack/lib/constants.sh to include cmd/distribution in the cross-compile targets
  - Vendored dependencies: Added required dependencies for the generic registry including spf13/cobra, bugsnag, and autocert packages

The openshift appliance will be replacing its use of an internal registry distribution image with the generic distribution/distribution binary from the image-registry image.


images/registry/config.yaml
zaneb: Not opposed to having this, but I kind of figured we would bind-mount the config into the container so that all the various configuration could be changed in one place, rather than using a default config from the container image.

rwsu:
/test e2e-metal-ipi-ovn-ipv6
conformance test failures

PR #452, opened 2025-11-20
CNTRLPLANE-1797: bump kubernetes to v1.34.2

ricardomaraschini:
bump kubernetes and related dependencies to the latest version.


ricardomaraschini:
> do you think we need to wait until the operator bump is ready before merging this one?

I don't see why. Should we ?

PR #465, opened 2026-02-17
[release-4.17] OCPBUGS-68182: update logrus to v1.9.1

dusk125:
I chose to do the minimum upgrade to fix the issue -> v1.9.1, but it does seem like this could safely be bumped to v1.9.3 if that's preferred.


flavianmissi:
/retitle [release-4.17] OCPBUGS-68182: update logrus to v1.9.1
/lgtm

PR #468, opened 2026-04-29
OCPBUGS-84680: pin google cloud storage sdk to v1.30.1 to fix blob deletion

ricardomaraschini:
The library-go bump in commit fcb7f1f97 upgraded cloud.google.com/go/storage from v1.30.1 to v1.56.0 as a transitive dependency. This upgrade included a breaking change in v1.55.0 where the auth library was migrated from golang.org/x/oauth2 (google.Credentials) to cloud.google.com/go/auth (auth.Credentials).

This auth library migration broke blob deletion operations on GCS storage, causing HTTP 500 errors during image pruning operations. The auth.Credentials type does not properly apply storage.ScopeFullControl to Delete API calls, even though the scope is correctly requested.

This change pins the Google Cloud packages to the last known working versions from 4.21:
- cloud.google.com/go v0.110.6 (was v0.121.6)
- cloud.google.com/go/storage v1.30.1 (was v1.56.0)
- cloud.google.com/go/iam v1.1.1 (was v1.5.3)
- cloud.google.com/go/compute/metadata v0.3.0 (was v0.9.0)

The library-go bump and TLS cipher configuration features remain intact, only the Google Cloud SDK versions are overridden via replace directives.

## Summary by CodeRabbit

* **Chores**
  * Reduced indirect dependency footprint by removing several unused transitive modules to streamline builds.
  * Pinned the Cloud Storage client to v1.30.1 to ensure a consistent, stable runtime version.


ricardomaraschini:
closing.

PR #470, opened 2026-04-30
OCPBUGS-84738: pin cloud.google.com/go/storage to v1.30.1 to fix blob deletion

ricardomaraschini:
**This is a manual cherry-pick of https://github.com/openshift/image-registry/pull/469**

The library-go bump in commit fcb7f1f97 upgraded cloud.google.com/go/storage from v1.30.1 to v1.56.0 as a transitive dependency. This upgrade included a breaking change in v1.55.0 where the auth library was migrated from golang.org/x/oauth2 (google.Credentials) to cloud.google.com/go/auth (auth.Credentials).

This auth library migration broke blob deletion operations on GCS storage, causing HTTP 500 errors during "oc adm prune images" operations. The new auth.Credentials type does not properly apply storage.ScopeFullControl to Delete API calls, even though the scope is correctly requested in the code.

Evidence:
- Storage v1.30.1 uses google.Credentials (works correctly)
- Storage v1.56.0 uses auth.Credentials (broken - scopes not applied)
- Delete code unchanged between versions
- HTTP client implementation identical
- Only difference: credentials type

This change pins cloud.google.com/go/storage to v1.30.1 (the last known working version from 4.21) via a replace directive. Other Google Cloud packages remain at their current versions and are compatible.

The library-go bump and TLS cipher configuration features remain intact.


ricardomaraschini:
I manually verified this.

/verified by @ricardomaraschini

PR #475, opened 2026-07-22
OCPBUGS-99398: Fix resolveUpstreamRef to properly propagate not-found errors

akesarka:
## Summary

- Fix resolveUpstreamRef() to use convertImageStreamGetterError() instead of hardcoding ErrImageStreamUnknownErrorCode, so that not-found errors from imageStreamGetter.layers() are properly converted to ErrImageStreamNotFoundCode.
- This fixes HTTP 500 responses on manifest HEAD requests for non-existent ImageStreams, which broke ImageStream auto-creation with Docker Engine 29+ clients.

## Problem

Docker Engine 29+ (with containerd image store) wraps single-platform images in OCI indexes and issues HEAD requests to check sub-manifest existence before PUT. When the ImageStream does not exist, manifestService.Exists() -> GetImageOfImageStream() -> resolveUpstreamRef() returns ErrImageStreamUnknownErrorCode instead of ErrImageStreamNotFoundCode, causing the registry to respond with HTTP 500 instead of 404.

This prevents the Docker client from proceeding to the manifest PUT, which is where CreateImageStreamMapping() auto-provisions the ImageStream.

## Root Cause

resolveUpstreamRef() was the only method in pkg/imagestream/imagestream.go that did not use convertImageStreamGetterError() to translate getter errors. All other methods (ResolveImageID, TagIsInsecure, Exists, localRegistry, IdentifyCandidateRepositories, Tags) already use this pattern.

## Test plan

- [ ] Verify the fix compiles and passes existing unit tests
- [ ] Push an image with Docker 29+ to a non-existent ImageStream and confirm auto-creation works
- [ ] Confirm existing pull-through and manifest operations are unaffected

Fixes: https://redhat.atlassian.net/browse/OCPBUGS-99398

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved error reporting when image streams or image layers cannot be retrieved, providing more accurate error classification and preserving the original error message.

### DIFF
--- a/pkg/imagestream/imagestream.go
+++ b/pkg/imagestream/imagestream.go
@@ -233,10 +233,9 @@ func (is *imageStream) GetImageOfImageStream(ctx context.Context, dgst digest.Di
 func (is *imageStream) resolveUpstreamRef(ctx context.Context, dgst digest.Digest) (reference.DockerImageReference, rerrors.Error) {
 	layers, rErr := is.imageStreamGetter.layers()
 	if rErr != nil {
-		return reference.DockerImageReference{}, rerrors.NewError(
-			ErrImageStreamUnknownErrorCode,
-			fmt.Sprintf("resolveUpstreamRef: failed to get layers for image stream %s", is.Reference()),
+		return reference.DockerImageReference{}, convertImageStreamGetterError(
 			rErr,
+			fmt.Sprintf("resolveUpstreamRef: failed to get layers for image stream %s", is.Reference()),
 		)
 	}
 
--- a/pkg/imagestream/imagestream.go
+++ b/pkg/imagestream/imagestream.go
@@ -199,6 +199,13 @@ func (is *imageStream) getImageOfImageStream(ctx context.Context, dgst digest.Di
 // not be sent to the master API. If you need unmodified version of the
 // image object, please use getStoredImageOfImageStream.
 func (is *imageStream) GetImageOfImageStream(ctx context.Context, dgst digest.Digest) (*imageapiv1.Image, rerrors.Error) {
+	if _, rErr := is.imageStreamGetter.get(); rErr != nil {
+		return nil, convertImageStreamGetterError(
+			rErr,
+			fmt.Sprintf("GetImageOfImageStream: image stream %s not found", is.Reference()),
+		)
+	}
+
 	isImage, err := is.getImageOfImageStream(ctx, dgst)
 	if err == nil {
 		return isImage, nil
--- a/pkg/imagestream/imagestream_error_test.go
+++ b/pkg/imagestream/imagestream_error_test.go
@@ -1,0 +1,46 @@
+package imagestream
+
+import (
+	"testing"
+
+	rerrors "github.com/openshift/image-registry/pkg/errors"
+)
+
+func TestConvertImageStreamGetterError(t *testing.T) {
+	tests := []struct {
+		name         string
+		inputCode    string
+		expectedCode string
+	}{
+		{
+			name:         "NotFound error is propagated as ImageStream NotFound",
+			inputCode:    ErrImageStreamGetterNotFoundCode,
+			expectedCode: ErrImageStreamNotFoundCode,
+		},
+		{
+			name:         "Forbidden error is propagated as ImageStream Forbidden",
+			inputCode:    ErrImageStreamGetterForbiddenCode,
+			expectedCode: ErrImageStreamForbiddenCode,
+		},
+		{
+			name:         "Unknown error remains as ImageStream Unknown",
+			inputCode:    ErrImageStreamGetterUnknownCode,
+			expectedCode: ErrImageStreamUnknownErrorCode,
+		},
+	}
+
+	for _, tc := range tests {
+		t.Run(tc.name, func(t *testing.T) {
+			inputErr := rerrors.NewError(tc.inputCode, "test/repo", nil)
+			result := convertImageStreamGetterError(inputErr, "test message")
+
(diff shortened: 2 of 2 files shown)


akesarka:
@flavianmissi @ricardomaraschini 

Could you please spare some time to review the request ?

pkg/imagestream/imagestream.go
```diff
@@ -233,10 +233,9 @@ func (is *imageStream) GetImageOfImageStream(ctx context.Context, dgst digest.Di
 func (is *imageStream) resolveUpstreamRef(ctx context.Context, dgst digest.Digest) (reference.DockerImageReference, rerrors.Error) {
 	layers, rErr := is.imageStreamGetter.layers()
 	if rErr != nil {
-		return reference.DockerImageReference{}, rerrors.NewError(
-			ErrImageStreamUnknownErrorCode,
-			fmt.Sprintf("resolveUpstreamRef: failed to get layers for image stream %s", is.Reference()),
+		return reference.DockerImageReference{}, convertImageStreamGetterError(
 			rErr,
+			fmt.Sprintf("resolveUpstreamRef: failed to get layers for image stream %s", is.Reference()),
```
ricardomaraschini, line 245: This seems to be correct. Can we have a unit test in place ?

ricardomaraschini (commented):
The problem being addressed here seems to be related to missing "layers" inside an already ImageStream, no ? Would you mind clarifying or even point me out to an upstream change that caused this change in behavior ?

akesarka:
> The problem being addressed here seems to be related to missing "layers" inside an already ImageStream, no ? Would you mind clarifying or even point me out to an upstream change that caused this change in behavior ?

The issue is not about missing layers inside an existing ImageStream — it's about the ImageStream itself not existing yet during a push.

Docker Engine 29+ now uses the containerd image store by default, which wraps single-platform images in an OCI image index. During push, the Docker client issues a HEAD /v2/<namespace>/<repo>/manifests/sha256:<digest> request to check if the sub-manifest already exists before issuing the manifest PUT.

When the ImageStream doesn't exist, manifestService.Exists() calls GetImageOfImageStream(), which first tries getImageOfImageStream() (fails — IS not found via get()), then falls through to resolveUpstreamRef(). Inside resolveUpstreamRef(), imageStreamGetter.layers() also returns ErrImageStreamGetterNotFoundCode, but this was being wrapped as ErrImageStreamUnknownErrorCode instead of being properly converted. This caused Exists() to return a generic error (HTTP 500) instead of distribution.ErrBlobUnknown (HTTP 404).

The Docker client treats 500 as fatal and aborts — the manifest PUT (where CreateImageStreamMapping() auto-provisions the ImageStream) is never reached.

Upstream change that triggers this: [moby/moby#51532](https://github.com/moby/moby/issues/51532) — Docker 29 enabled containerd image store by default, changing push behavior.

ricardomaraschini (commented):
> The issue is not about missing layers inside an existing ImageStream

This seems to contradict what comes next:

> When the ImageStream doesn't exist, manifestService.Exists() calls GetImageOfImageStream(), which first tries getImageOfImageStream() (fails — IS not found via get()), then falls through to resolveUpstreamRef(). Inside resolveUpstreamRef(), imageStreamGetter.layers()

As you can see the problem happens when listing the "layers". At that stage the ImageStream should be there already. That is the missing information. 

Can you please paste here what are the resulting ImageStream/Images created after pushing with Docker v29 (with this fix in place)? The YAMLS would be ideal.

akesarka:
@ricardomaraschini 

Added a detailed [reproduction document ](https://docs.google.com/document/d/1de5ZpijKqbAKa3gIhAcHkzjsdOzfJCQX4bvbnXZ1qIg/edit?tab=t.0#heading=h.ht7sc0vcwj74) demonstrating the bug on  OCP 4.20.24 cluster with Docker CE 29.4.0 + containerd image store:

Key findings:

- docker buildx build --push to a non-existent ImageStream fails with repeated 500 Internal Server Error on HEAD /manifests/sha256:...
- podman push and legacy docker push to the same non-existent ImageStream succeed (they don't use HEAD-by-digest)
- Registry logs confirm: err.code=unknown err.detail="ImageStream:Unknown: resolveUpstreamRef: ... ImageStreamGetter:NotFound: ..."

ricardomaraschini:
/retitle OCPBUGS-99398: Fix resolveUpstreamRef to properly propagate not-found errors

pkg/imagestream/imagestream.go
ricardomaraschini, line 245: We should verify if the ImageStream exists in `GetImageOfImageStream()` before doing anything else. Right now, if `getImageOfImageStream` fails, the code falls through to `resolveUpstreamRef` but it should never reach that point if the ImageStream does not exist.

Maybe add an early call to `is.imageStreamGetter.get()` (right at the start of `GetImageOfImageStream`). If it returns an error, wrap it with `convertImageStreamGetterError` and return immediately. This would avoid extra API calls.

You may even keep the current change as a safety net, just in case.

ricardomaraschini (changes requested):

pkg/imagestream/imagestream.go
akesarka: Updated as per your suggestion — added an early ImageStream existence check at the top of GetImageOfImageStream():

Could you please have a look if that is what you asked for ?

The original fix in resolveUpstreamRef is retained as a safety net for any other code paths that call it directly.

ricardomaraschini:
@gangwgr Would you be able to verify this one ?

